PR #178, opened 2026-06-24
feat(mcp): declare context7 + deepwiki as user-scope MCP servers

kryota-dev:
## Summary

Migrates the two globally-useful MCP servers — **context7** (library docs) and **deepwiki** (repo Q&A) — from this repo's project-scoped `.mcp.json` to **user scope**, so they are available in every project instead of only inside `dotfiles`.

This is **PR7** of the Phase 6 stack. Scope was intentionally narrowed to **task #33 only**; the tool-install tasks #16/#17/#18 (AgentShield / firecrawl+Exa MCP / dmux) are **deferred** to ride with the ECC skill-external work (#46/#15), because their dependent ECC skills (`security-scan` / `deep-research` / `exa-search` / `dmux-workflows`) are not deployed yet and #17 additionally requires firecrawl/Exa API keys.

## Changes

- **`home/run_onchange_after_13-setup-mcp.sh.tmpl`** (new): a `run_onchange` lifecycle script that idempotently registers context7 + deepwiki as user-scope servers for **every account config dir** (`~/.claude` and `~/.claude-r06`). It drives `claude mcp add-json --scope user` (remove-then-add), is fail-soft, and skips when the `claude` CLI is absent. `~/.claude.json`-style user config is volatile (rewritten by the CLI per session), so it is driven via the CLI rather than owned by chezmoi.
- **`.mcp.json`**: dropped context7/deepwiki; kept the project-specific **spec-workflow** (this repo has an active `.spec-workflow/`, so per the design principle "project-specific servers stay in the repo that uses them", spec-workflow is retained — this deviates from the plan's literal "git rm .mcp.json", verified against the primary source).
- **`tests/files.bats`**: added coverage for the setup script (declares both servers with `--scope user`) and the `.mcp.json` contract (keeps spec-workflow, no longer declares context7/deepwiki).

## Verification (mine)

- `make lint` ✅ / `make test` ✅ (68 bats, incl. 2 new)
- `chezmoi execute-template` renders the script cleanly; `bash -n` OK; chezmoi recognizes `13-setup-mcp.sh`
- End-to-end against a sandbox config dir: both servers register as **"User config (available in all your projects)"**, idempotent re-run does not error

## Runtime checklist (user)

- [ ] `chezmoi apply` → `13-setup-mcp.sh` runs
- [ ] `cld` → `claude mcp list` shows context7 + deepwiki as **user** scope
- [ ] `cld-r06` → `claude mcp list` shows context7 + deepwiki as **user** scope
- [ ] In `dotfiles`, spec-workflow still resolves (project scope); context7/deepwiki now come from user scope

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,15 +1,5 @@
 {
   "mcpServers": {
-    "context7": {
-      "type": "stdio",
-      "command": "npx",
-      "args": ["-y", "@upstash/context7-mcp"],
-      "env": {}
-    },
-    "deepwiki": {
-      "command": "npx",
-      "args": ["mcp-remote", "https://mcp.deepwiki.com/mcp"]
-    },
     "spec-workflow": {
       "command": "npx",
       "args": ["-y", "@pimzino/spec-workflow-mcp@latest", "."]
--- a/home/run_onchange_after_13-setup-mcp.sh.tmpl
+++ b/home/run_onchange_after_13-setup-mcp.sh.tmpl
@@ -1,0 +1,47 @@
+#!/bin/bash
+set -euo pipefail
+
+# Declaratively register user-scope MCP servers for every Claude Code account.
+#
+# `claude mcp add-json --scope user` writes to "$CLAUDE_CONFIG_DIR/.claude.json", which the
+# CLI rewrites on every session (volatile). chezmoi must not own that file directly, so this
+# script drives the CLI instead. run_onchange re-runs whenever the server definitions below
+# change (the rendered script body is the change key).
+#
+# Only globally-useful servers belong here: context7 (library docs) and deepwiki (repo Q&A)
+# are wanted in every project. Project-specific servers (spec-workflow, serena) stay in each
+# repo's own .mcp.json. claude.ai connectors and the Chrome extension are configured manually.
+
+if ! command -v claude &>/dev/null; then
+  echo "WARNING: claude CLI not found; skipping user-scope MCP setup."
+  exit 0
+fi
+
+# Each account is isolated by CLAUDE_CONFIG_DIR (see dot_config/zsh/claude.zsh: cld -> ~/.claude,
+# cld-r06 -> ~/.claude-r06). user-scope servers live in "<config dir>/.claude.json", so register
+# them once per account.
+config_dirs=("$HOME/.claude" "$HOME/.claude-r06")
+
+# Server name -> canonical JSON. Both are stdio (npx-launched); deepwiki bridges to its remote
+# HTTP endpoint via mcp-remote. Keep in sync with the project's .mcp.json removal of these two.
+mcp_names=("context7" "deepwiki")
+mcp_configs=(
+  '{"type":"stdio","command":"npx","args":["-y","@upstash/context7-mcp"]}'
+  '{"type":"stdio","command":"npx","args":["mcp-remote","https://mcp.deepwiki.com/mcp"]}'
+)
+
+for config_dir in "${config_dirs[@]}"; do
+  echo "==> Registering user-scope MCP servers in ${config_dir}..."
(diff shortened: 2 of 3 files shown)


kryota-dev (commented):
## 🤖 Multi-Review summary (PR #178)

Three independent reviewers — `cc-code-review`, `cc-security-review`, and `codex` (round 1 + round 2) — ran in parallel. All findings have been verified against primary sources and resolved; **PR is review-clean**.

### Overall

| Category | cc-code-review | cc-security-review | codex (r1) |
|---|---|---|---|
| MUST | 0 | 0 | 2 |
| SHOULD | 2 | 0 | 3 |
| NITS | 3 | — | 1 |
| Medium (sec) | — | 1 | — |
| GOOD | 5 | 7 | 1 |

**Security**: overall risk **Low**. The change runs entirely inside the trust boundary (applying one's own chezmoi config), takes no user input, and registers only hardcoded static JSON. No injection or secret-exposure path.

### [MUST] resolved (codex round 1 → confirmed fixed in round 2)

1. **PATH/ordering silent-skip** — `run_onchange_after_13` runs before `run_once_after_16` creates the `~/.local/bin/claude` launcher, so `command -v claude` was unreliable on a fresh apply; the old `exit 0` skip would have recorded success and never retried, leaving MCP servers unregistered. **Fix**: guard on `command -v mise` and invoke `mise exec -- claude` (mise + the mise-installed claude are both present by step 13, matching the `agent-browser` step-18 pattern). Verified: step-16 creates the symlink, step-12 runs `mise install --yes`, `mise exec -- claude --version` → `2.1.185`.
2. **Swallowed registration failure** — `add-json` failures were warned-and-ignored, so a partial state would be frozen as "success". **Fix**: aggregate failures and `exit 1` so chezmoi retries on the next apply.

### [SHOULD] resolved

- **deepwiki missing `-y`** (codex) — added `-y` to the `mcp-remote` npx args so the first run never blocks on an install prompt (parity with context7).
- **`~/.claude-r06` absence** (cc-code-review, codex) — **verified by primary source**: `claude mcp add-json --scope user` auto-creates the (even nested-absent) config dir and exits 0, so there is no silent failure. Also `~/.claude-r06` is always chezmoi-deployed. Added a defensive `mkdir -p` so the script no longer depends on that CLI behaviour.
- **grep-only tests** (cc-code-review, codex) — replaced with a fake-`mise` execution test that asserts both servers are `add-json`'d with `--scope user` for **both** account config dirs, plus a `jq` validity check of the server JSON.

### Verified-and-declined (kept intentionally)

- **[Medium / cc-security] unpinned npx packages** — `@upstash/context7-mcp` / `mcp-remote` are not version-pinned. This is **inherited verbatim from the existing `.mcp.json`** (risk delta ≈ 0 — the definitions are moved, not introduced), and the repo/upstream convention is unpinned (`spec-workflow` itself uses `@latest`). Pinning would deviate from upstream install docs and add bump maintenance. Out of scope for this scope-migration PR; can be revisited as a dedicated supply-chain pass.

### [GOOD] (highlights)

- Hardcoded static JSON → no JSON/command injection path; all expansions quoted; `set -euo pipefail`.
- CLI-driven registration (not chezmoi-owned) is the right call for the volatile `.claude.json`.
- Account isolation via `CLAUDE_CONFIG_DIR` matches `dot_config/zsh/claude.zsh` (cld / cld-r06).
- No secrets; CLI output suppressed; negative assertions in tests catch config drift.

### My verification

`make lint` ✅ · `make test` ✅ (69 bats) · `chezmoi execute-template` + `bash -n` + `shellcheck` ✅ · end-to-end against a sandbox config dir: both servers register as **User config (available in all your projects)**, idempotent re-run clean.

---
*Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>*

PR #249, opened 2026-07-06
docs(sdd): drop deprecated TeamCreate/TeamDelete and team_name from review phase

kryota-dev:
## Summary

Claude Code v2.1.178+ removed the `TeamCreate` / `TeamDelete` tools, and the
`Agent` tool's `team_name` field is now accepted but ignored. This updates the
SDD skill's review phase (Phase 5) to match the current Agent tool spec.

## Changes

- **Remove the `TeamCreate` step** (old §5-2). Review agents are spawned by
  calling the `Agent` tool directly — no team setup is required.
- **Remove the `team_name` field** from the review-agent spawn example and from
  the in-prompt team-info block (SendMessage `recipient`-based coordination is
  kept, as it is still supported).
- **Remove the `TeamDelete` step** (old §5-7). Spawned agents are auto-cleaned
  up at session end, so explicit team deletion is unnecessary.
- **Renumber Phase 5 subsections** to 5-1..5-5 after dropping the two steps.

Docs-only change to `home/dot_agents/skills/sdd/SKILL.md`; no shell/CI/config
files affected.

## Verification

- `grep -rn -E 'TeamCreate|TeamDelete|team_name' home/` → no matches
- Phase 5 headers are sequential 5-1..5-5

Ref: https://code.claude.com/docs/en/agent-teams.md


home/dot_agents/skills/sdd/SKILL.md
```diff
@@ -316,25 +316,16 @@ tasks.md の各タスクを順番に実装:
...
-```
-TeamCreate:
-  team_name: "sdd-review-{spec-name}"
-  description: "Code Review Team for {spec-name}"
-```
-
-### 5-3. レビューエージェントのスポーン
-
-考察した構成に基づき、各レビューエージェントを **Agent ツール** でスポーン。
+考察した構成に基づき、各レビューエージェントを **Agent ツール** でスポーン。事前のチーム作成は不要（Agent ツールを直接呼び出すだけでよく、スポーンしたエージェントはセッション終了時に自動クリーンアップされる）。
```
kryota-dev, line 321: [SHOULD] The current wording matches the official doc ("cleanup happens automatically when the session exits"), so it is accurate. Optional clarity improvement: since the per-agent shutdown flow is retained in section 5-4, you could phrase the automatic cleanup as the session-end backstop while explicit `TeamDelete` is what is no longer needed — so it does not read as if per-agent shutdown is unnecessary. Not a blocker.

Note: codex proposed rewording this to say team state is auto-created on first teammate spawn; that mechanism is not stated in the official doc, so it was not adopted.

home/dot_agents/skills/sdd/SKILL.md
```diff
@@ -374,8 +365,7 @@ Agent:
     ### 良い点
     実装の優れている点を挙げる
 
-    ## チーム情報
-    - チーム名: sdd-review-{spec-name}
+    ## 連絡情報
```
kryota-dev, line 368: [NITS] `## 連絡情報` reads a little literal. Since this block lists addressing info (the teammate's own name and the lead's recipient name), a heading like `## 宛先情報` may convey the intent more clearly. Optional.

kryota-dev (commented):
## Multi-Review — integrated result (PR #249)

Docs-only change to `home/dot_agents/skills/sdd/SKILL.md` (Phase 5). Three reviewers ran in parallel (cc-code-review, cc-security-review, codex). All premises were cross-checked against the official Agent Teams doc (https://code.claude.com/docs/en/agent-teams.md).

### Verdict
**No MUST — mergeable.** Removing `TeamCreate` / `TeamDelete` / `team_name`, the 5-1..5-5 renumbering, and the retained `SendMessage` / `recipient` coordination all match the current Agent tool spec (verified against the official doc).

### Totals
|---|---|---|---|
| MUST | 0 | 0 | 0 |
| SHOULD | 1 | 0 | 1 |
| NITS | 2 | 0 | 1 |
| GOOD | 2 | — | 2 |

### Security
Overall risk: **Low**, 0 vulnerabilities. One Info questioned the accuracy of the "auto-cleanup at session exit" wording — verified against the official doc, which states cleanup happens automatically when the session exits and there is no separate cleanup step. Wording confirmed accurate.

### In-scope inline comments
- [SHOULD] L321 / L490 — auto-cleanup wording clarity (non-blocking).
- [NITS] L368 — heading naming.

### Fact-check ledger (dismissed / out-of-scope)
- **[SHOULD, cc-code-review] `team-lead` recipient name may not resolve to the lead's address** — pre-existing (context line, not changed by this PR), low confidence / unverified against the doc. Out of scope for this migration.
- **[NITS, cc-code-review] Phase 5 lacks a note that agent teams are experimental (`CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS`)** — pre-existing, not introduced here. Out of scope.
- **[SHOULD, codex] Proposed rewrite** claims "team state is auto-created on first teammate spawn" — this mechanism is **not** stated in the official doc (which only says spawning no longer needs a setup step). Declined the rewrite to avoid adding an unverified claim; the current wording already matches the doc. Downgraded to the clarity note on L321.

### GOOD
- No residual `TeamCreate` / `TeamDelete` / `team_name`; renumbering and cross-references are correct.
- Correctly retained `SendMessage` / `recipient` / `shutdown_request` coordination (still part of the current spec — removing it would have broken the flow).

home/dot_agents/skills/sdd/SKILL.md
kryota-dev: **No change — verified accurate.**

The current wording is a faithful paraphrase of the official doc ("cleanup happens automatically when the session exits" / "there's no separate cleanup step"). §5-4 still retains the per-agent `shutdown_request` flow and §5-5 states that explicit `TeamDelete` is unnecessary only after shutdown, so the text is internally consistent and does not imply per-agent shutdown is unnecessary. Keeping as-is to avoid adding the unverified "team state auto-created on first spawn" mechanism codex proposed.

home/dot_agents/skills/sdd/SKILL.md
kryota-dev: **No change — optional style only.**

`## 連絡情報` is acceptable and unambiguous in context. Renaming to `## 宛先情報` is a purely cosmetic preference and not worth a separate commit + full CI cycle (KISS/YAGNI). Leaving as-is; can be folded into a future docs polish if desired.

PR #291, opened 2026-07-09
fix(claude): inline fable orchestrator prompt for happy variants (hcldf/hcldf-r06)

kryota-dev:
## Summary

`hcldf` / `hcldf-r06` (the Fable 5 orchestrator variants routed through the `happy` wrapper) abort at launch on Claude Code v2.1.185+ with:

```
Error: Cannot use both --append-system-prompt and --append-system-prompt-file. Please use only one.
```

**Root cause:** the `happy` (slopus/happy) wrapper always injects its own `--append-system-prompt` before appending the caller's args ([`happy` dist source](https://github.com/slopus/happy): `args.push("--append-system-prompt", systemPrompt)`). `_claude_fable` additionally passed the orchestrator prompt via `--append-system-prompt-file`. Claude Code >= 2.1.185 newly rejects mixing the two flag forms, so every happy-routed Fable launch failed. The direct `cldf` / `cldf-r06` variants were unaffected (no happy = no injected `--append-system-prompt`), and the non-Fable `hcld` / `hcld-r06` variants were unaffected too (they never add the prompt-file flag).

## Fix

In `_claude_fable`, when routing through `happy` (`$1 == happy`), inline the orchestrator prompt via `--append-system-prompt` (content) instead of `--append-system-prompt-file`. Repeating the same flag is allowed, so ours coexists with happy's injected one. The direct variants keep `--append-system-prompt-file` to stay out of argv. The prompt is small (~6 KB) and static, so the argv cost of inlining is negligible.

## Verification

Confirmed empirically against `claude` 2.1.205:

- `--append-system-prompt` + `--append-system-prompt-file` → rejected at parse time (reproduces the reported error).
- `--append-system-prompt` specified twice → accepted, both prompts applied.

Test and lint:

- Added a regression test in `tests/zsh_aliases.bats` for the `happy` + readable-prompt-file path, asserting `--append-system-prompt` (not `--append-system-prompt-file`) is used.
- `make lint` and `make test-bats` (156 tests) all pass.

## Deploy note

Run `chezmoi apply` after merge to restore `hcldf` / `hcldf-r06`.

### DIFF
--- a/home/dot_config/zsh/claude.zsh
+++ b/home/dot_config/zsh/claude.zsh
@@ -73,6 +73,12 @@ alias claude-config='ECC_DISABLED_HOOKS_EXTRA=pre:config-protection,pre:edit-wri
 # The prompt is passed via --append-system-prompt-file (path) instead of --append-system-prompt
 # (content) so the prompt body stays out of argv — the CLI reads the file at process start,
 # avoiding argv-length and control-char concerns as the prompt grows.
+# Exception (hcldf/hcldf-r06): the happy (slopus/happy) wrapper always injects its own
+# --append-system-prompt, and Claude Code >= 2.1.185 refuses to mix --append-system-prompt
+# with --append-system-prompt-file ("Cannot use both ... Please use only one."), which aborted
+# the happy variants at launch. So when routing through happy ($1 == happy) the prompt is
+# inlined via --append-system-prompt instead — repeating the same flag is allowed, so ours
+# coexists with happy's. The prompt is small (~6 KB) and static, so the argv cost is moot.
 _claude_fable() {
   local home_dir="$1"
   shift
@@ -81,7 +87,13 @@ _claude_fable() {
   (($#)) || set -- claude
   local prompt_file="$HOME/.claude/fable-orchestrator-prompt.md"
   local -a fable_flags=(--model claude-fable-5)
-  [[ -r "$prompt_file" ]] && fable_flags+=(--append-system-prompt-file "$prompt_file")
+  if [[ -r "$prompt_file" ]]; then
+    if [[ "$1" == happy ]]; then
+      fable_flags+=(--append-system-prompt "$(<"$prompt_file")")
+    else
+      fable_flags+=(--append-system-prompt-file "$prompt_file")
+    fi
+  fi
   _claude_with_home "$home_dir" "$@" "${fable_flags[@]}"
 }
 alias cldf='_claude_fable "$HOME/.claude" claude'
--- a/home/dot_config/zsh/claude.zsh
+++ b/home/dot_config/zsh/claude.zsh
@@ -77,8 +77,11 @@ alias claude-config='ECC_DISABLED_HOOKS_EXTRA=pre:config-protection,pre:edit-wri
 # --append-system-prompt, and Claude Code >= 2.1.185 refuses to mix --append-system-prompt
 # with --append-system-prompt-file ("Cannot use both ... Please use only one."), which aborted
 # the happy variants at launch. So when routing through happy ($1 == happy) the prompt is
-# inlined via --append-system-prompt instead — repeating the same flag is allowed, so ours
-# coexists with happy's. The prompt is small (~6 KB) and static, so the argv cost is moot.
+# inlined via --append-system-prompt instead — repeating the same flag is allowed (observed on
+# claude 2.1.205), so ours coexists with happy's. The prompt is small (~6 KB) and static, so the
+# argv cost is moot; note that inlining exposes the body in argv (readable by same-user
+# processes), so keep fable-orchestrator-prompt.md secret-free — revert to the file form if
+# happy ever honors --append-system-prompt-file.
 _claude_fable() {
   local home_dir="$1"
   shift
--- a/tests/zsh_aliases.bats
+++ b/tests/zsh_aliases.bats
@@ -95,6 +95,27 @@ load helpers/setup
   [ "$output" = "happy|claude --resume --model claude-fable-5" ]
 }
 
+@test "claude.zsh: _claude_fable inlines the orchestrator prompt via --append-system-prompt through happy" {
+  # Regression guard for hcldf/hcldf-r06: the happy (slopus/happy) wrapper always injects its
+  # own --append-system-prompt, and Claude Code >= 2.1.185 rejects mixing --append-system-prompt
+  # with --append-system-prompt-file ("Cannot use both ... Please use only one."). So when
+  # routing through happy, the orchestrator prompt must be inlined via --append-system-prompt
+  # (repeating the same flag is allowed) instead of --append-system-prompt-file, which would
+  # otherwise collide with happy's own flag and abort the launch. Direct (non-happy) launches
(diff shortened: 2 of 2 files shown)


kryota-dev (commented):
## Multi-Review — integrated result (PR #291)

Second opinion from three independent reviewers (`cc-code-review`, `cc-security-review`, `codex`/gpt-5.5). No pre-existing reviews to dedup against.

**Verdict: LGTM — no blocking (MUST) issues.** The fix is correct: `happy` always injects its own `--append-system-prompt`, so inlining the orchestrator prompt via `--append-system-prompt` (a repeated flag, which is allowed) is the right way to avoid the rejected mix with `--append-system-prompt-file`.

|---|---|---|---|
| MUST | 0 | 0 | 0 |
| SHOULD | 1 | Low: 1 | 1 |
| NITS | 2 | — | 2 |
| GOOD | 5 | overall risk Low | 2 |

**Security:** overall risk **Low**. One Medium note: inlining moves the ~5.6 KB prompt body into argv, which same-user processes can read on macOS (`KERN_PROCARGS2`; `ps` shows a truncated form). The current prompt contains no secrets (verified), so present risk is zero — the note is future-proofing.

**Consensus SHOULD (all three):** the regression test uses a single-word fixture; a multi-line/whitespace fixture would additionally guard the fix's core guarantee — that the prompt body survives as a single argv element.

**Fact-check (verified against primary sources before posting):**
- Repeating `--append-system-prompt` is accepted by `claude` 2.1.205 — reproduced (two flags → exit 0, both applied); the mix with `--append-system-prompt-file` is rejected at parse time (reproduces the reported error).
- `$(<file)` strips the trailing newline in zsh and preserves internal newlines as a single argument — reproduced.

The inline comments below carry the actionable items; they will be addressed in a follow-up commit on this branch.

tests/zsh_aliases.bats
```diff
@@ -95,6 +95,27 @@ load helpers/setup
...
+@test "claude.zsh: _claude_fable inlines the orchestrator prompt via --append-system-prompt through happy" {
+  # Regression guard for hcldf/hcldf-r06: the happy (slopus/happy) wrapper always injects its
+  # own --append-system-prompt, and Claude Code >= 2.1.185 rejects mixing --append-system-prompt
+  # with --append-system-prompt-file ("Cannot use both ... Please use only one."). So when
+  # routing through happy, the orchestrator prompt must be inlined via --append-system-prompt
+  # (repeating the same flag is allowed) instead of --append-system-prompt-file, which would
+  # otherwise collide with happy's own flag and abort the launch. Direct (non-happy) launches
+  # keep --append-system-prompt-file (covered by the readable-prompt-file test above).
+  mkdir -p "$BATS_TEST_TMPDIR/.claude"
+  printf 'ORCHESTRATOR-BODY\n' >"$BATS_TEST_TMPDIR/.claude/fable-orchestrator-prompt.md"
```
kryota-dev: [SHOULD] All three reviewers independently flagged this: the fixture is a single word (`ORCHESTRATOR-BODY`), so it does not guard the fix's core guarantee — that a real multi-line/whitespace prompt body survives as one argv element. Suggest a multi-line fixture plus an arg-count / exact-arg assertion, e.g.:

```bash
printf 'line one\nline two with spaces\n' >"$BATS_TEST_TMPDIR/.claude/fable-orchestrator-prompt.md"
happy() { print -r -- "argc=$#"; print -r -- "arg5=[$5]"; }
# argc=5, arg4=--append-system-prompt, arg5=[line one<newline>line two with spaces] (trailing newline stripped)
```

Verified locally: in zsh `$(<file)` strips the trailing newline and keeps the internal newline as a single argument, so this assertion is stable.

home/dot_config/zsh/claude.zsh
```diff
@@ -81,7 +87,13 @@ _claude_fable() {
   (($#)) || set -- claude
   local prompt_file="$HOME/.claude/fable-orchestrator-prompt.md"
   local -a fable_flags=(--model claude-fable-5)
-  [[ -r "$prompt_file" ]] && fable_flags+=(--append-system-prompt-file "$prompt_file")
+  if [[ -r "$prompt_file" ]]; then
+    if [[ "$1" == happy ]]; then
+      fable_flags+=(--append-system-prompt "$(<"$prompt_file")")
```
kryota-dev, line 95: [SHOULD] (security — Low / defensive) Inlining moves the prompt body into argv, which same-user processes can read on macOS (`KERN_PROCARGS2`; `ps` shows a truncated form). The current orchestrator prompt has no secrets, so risk is currently zero — this is future-proofing. Worth a one-line note (near this code, or at the top of `fable-orchestrator-prompt.md`) that the file's content becomes argv-visible under `hcldf`/`hcldf-r06` and must stay secret-free. If `happy` ever honors `--append-system-prompt-file`, revert to the path form to eliminate the exposure.

home/dot_config/zsh/claude.zsh
```diff
@@ -73,6 +73,12 @@ alias claude-config='ECC_DISABLED_HOOKS_EXTRA=pre:config-protection,pre:edit-wri
 # The prompt is passed via --append-system-prompt-file (path) instead of --append-system-prompt
 # (content) so the prompt body stays out of argv — the CLI reads the file at process start,
 # avoiding argv-length and control-char concerns as the prompt grows.
+# Exception (hcldf/hcldf-r06): the happy (slopus/happy) wrapper always injects its own
+# --append-system-prompt, and Claude Code >= 2.1.185 refuses to mix --append-system-prompt
+# with --append-system-prompt-file ("Cannot use both ... Please use only one."), which aborted
+# the happy variants at launch. So when routing through happy ($1 == happy) the prompt is
+# inlined via --append-system-prompt instead — repeating the same flag is allowed, so ours
```
kryota-dev: [NITS] "repeating the same flag is allowed" is observed behavior, not a documented CLI contract. Worth noting it was measured on `claude` 2.1.205, so a future reader can tell spec from observation.

tests/zsh_aliases.bats
kryota-dev: Addressed in 355faaa: the fixture is now multi-line with internal whitespace (`line one` / `line two with spaces`), and the happy mock asserts arg count == 5, `$4 == --append-system-prompt`, and `$5` equals the exact body (trailing newline stripped by `$(<file)`). This guards the single-argv-element boundary the single-word fixture missed. Verified locally: `make lint` + full bats (156) pass.

home/dot_config/zsh/claude.zsh
kryota-dev: Addressed in 355faaa (comment): noted that inlining exposes the body in argv (same-user readable on macOS) so `fable-orchestrator-prompt.md` must stay secret-free, and to revert to the file form if happy ever honors `--append-system-prompt-file`. The current prompt has no secrets, so present risk is zero — keeping this as a documented constraint rather than a behavioral change.

home/dot_config/zsh/claude.zsh
kryota-dev: Addressed in 355faaa: the comment now states the repeated-flag behavior was observed on `claude` 2.1.205, distinguishing observation from documented CLI contract.

PR #292, opened 2026-07-09
feat(claude): manage Claude Code plugins declaratively from settings.json

kryota-dev:
## Summary

Enables the Anthropic-official `claude-code-setup` plugin, and makes the whole plugin set
reproducible from `chezmoi apply` alone.

`settings.json` already declared `codex@openai-codex` under `enabledPlugins`, but that declaration
never installed anything. The work account (`~/.claude-r06`) was missing the codex plugin entirely,
because its marketplace was never registered there. This PR fixes the class of bug, not just the
instance.

## Why the declaration alone is not enough

Plugin state is split across two places:

| State | Lives in | chezmoi |
|---|---|---|
| Declaration | `settings.json`: `enabledPlugins`, `extraKnownMarketplaces` | owned, versioned |
| Runtime | `$CLAUDE_CONFIG_DIR/plugins/` (marketplace clones, install cache) | `.chezmoiignore`d — empty on a new machine |

The CLI does not bridge them. `enabledPlugins` only toggles a plugin that is *already* installed, and
declaring a marketplace does not register it. Both are upstream wontfix:

- [anthropics/claude-code#23737](https://github.com/anthropics/claude-code/issues/23737) — closed as duplicate; states `enabledPlugins` does not auto-install.
- [anthropics/claude-code#45323](https://github.com/anthropics/claude-code/issues/45323) — closed as not planned: "Plugins configured in org managed settings (`extraKnownMarketplaces` + `enabledPlugins`) do not auto-install in the CLI."

Verified against the version chezmoi actually installs — the mise-pinned **2.1.205**:
`enabledPlugins`, `extraKnownMarketplaces`, and `strictKnownMarketplaces` all appear in `strings`; the
`autoInstallEnabledPlugins` and `forcedPlugins` keys those issues ask for do not.

## Approach

`settings.json` stays the single source of truth. `run_onchange_after_17-setup-claude-plugins.sh.tmpl`
reads it at render time with `include | fromJson` and embeds just `enabledPlugins` and
`extraKnownMarketplaces` as JSON inside a **quoted heredoc**. So there is no second list to keep in
sync, no rendered value can be evaluated as shell source, and the `run_onchange` key flips only when
the declaration changes. The script then registers missing marketplaces and installs missing plugins,
once per account, exiting non-zero on failure so chezmoi retries on the next apply (the same contract
as `run_onchange_after_13-setup-mcp`).

### Pinning the marketplace

A marketplace is executable code that `chezmoi apply` installs unattended, so it must not track a
moving default branch. Measured in an isolated `CLAUDE_CONFIG_DIR`, pinning has two sharp edges:

| Method | Pinned? |
|---|---|
| `marketplace add openai/codex-plugin-cc#v1.0.4` | Yes — checks out `807e03a` |
| Declaring `ref` in `settings.json` only | **No — HEAD is checked out**, and the ref is not even recorded |
| `#<commit-sha>` | No — `fatal: Remote branch <sha> not found` (the ref reaches `git clone --branch`) |

So the `ref` is declared on the `github` source (the settings schema allows it) *and* rendered into
the CLI argument as `<repo>#<ref>`. Because a commit SHA cannot be used, the pin is a **tag**, which
is weaker than the commit pins this repo uses for chezmoi externals — a tag can be moved upstream.
Bumping it is a manual edit today; wiring Renovate's `github-tags` datasource is tracked separately.

Note that both accounts already have `openai-codex` registered from an earlier unpinned
`marketplace add`, and the script skips registration when the name is present. The pin therefore takes
effect on a fresh machine, or after the marketplace is removed and re-added.

### The settings.json rewrite hazard

`claude plugin install`, `claude plugin marketplace add`, and the interactive `/plugin` manager each
rewrite `settings.json` with their own serializer: top-level keys are reordered and every hook's `id`
and `description` annotation is dropped. Nothing breaks at runtime — hooks are identified by the
argument inside their `command`, not by those keys — but the annotations would be lost forever if the
rewritten file were ever `chezmoi add`ed. The script snapshots the file up front and writes it back on
exit, and repairs the work account's `settings.json` symlink if an atomic rewrite replaced it with a
regular file. The hazard is documented so a human doing the same by hand knows to run `chezmoi apply`.

## Changes

| File | Change |
|---|---|
| `home/dot_claude/settings.json` | Enable `claude-code-setup@claude-plugins-official`; declare the `openai-codex` marketplace, pinned to `v1.0.6` |
| `home/run_onchange_after_17-setup-claude-plugins.sh.tmpl` | New reconciler |
| `tests/files.bats` | Declaration invariants (both plugins enabled, every marketplace resolvable and pinned) plus three end-to-end tests driving the rendered script against fake `mise`/`claude` |
| `docs/agents/claude-code.{md,ja.md}` | Plugin set, why the declaration does not install, the pin's two sharp edges, the rewrite hazard |
| `docs/architecture/lifecycle-scripts.{md,ja.md}` | Script 17 in the timeline, tracked-external table, OS-guard table |

## Verification

- `make lint` — clean. Re-checked with a script that honours shellcheck's and shfmt's exit codes, because `make lint` currently swallows them (`Makefile:13-19`); every script under `home/` is clean either way.
- `make test` — 163 ok, 0 failures.
- The rendered script was run against the real environment: `~/.claude` was a no-op, `~/.claude-r06`
  gained the `openai-codex` marketplace and the `codex` plugin. `settings.json` came out byte-identical
  (same sha before and after), hook annotations intact, the work-account symlink intact. A second run
  finished in 0.67 s as a clean no-op.
- The three end-to-end bats tests were mutation-checked: deleting the `trap` line, or dropping the
  `#<ref>` suffix, each makes the corresponding assertion fail. They are not vacuous.

## Review

cc-code-review, cc-security-review, and codex reviewed this PR; every claim was checked against a
primary source before acting. Three MUSTs survived — a silent `exit 0` when `jq` is absent, rendering
settings.json values into bash array literals, and the unpinned marketplace — and all are fixed. See
the review thread for the findings that verification corrected or rejected.

## Checklist

- [x] Code review complete
- [x] Tests added
- [x] Configuration change impact verified (`chezmoi apply` reproducibility on a fresh machine)
- [x] Docs updated (EN canonical + JA mirror)

### DIFF
--- a/home/dot_claude/settings.json
+++ b/home/dot_claude/settings.json
@@ -473,8 +473,16 @@
   },
   "model": "claude-opus-4-8[1m]",
   "enabledPlugins": {
-    "codex@openai-codex": true
+    "codex@openai-codex": true,
+    "claude-code-setup@claude-plugins-official": true
   },
-  "extraKnownMarketplaces": {},
+  "extraKnownMarketplaces": {
+    "openai-codex": {
+      "source": {
+        "source": "github",
+        "repo": "openai/codex-plugin-cc"
+      }
+    }
+  },
   "agentPushNotifEnabled": true
 }
--- a/home/dot_claude/settings.json
+++ b/home/dot_claude/settings.json
@@ -480,7 +480,8 @@
     "openai-codex": {
       "source": {
         "source": "github",
-        "repo": "openai/codex-plugin-cc"
+        "repo": "openai/codex-plugin-cc",
+        "ref": "v1.0.6"
       }
     }
   },
--- a/home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
+++ b/home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
@@ -1,0 +1,148 @@
+#!/bin/bash
+set -euo pipefail
+
+# Reconcile each Claude Code account's plugin runtime with the declaration in dot_claude/settings.json.
+#
+# Plugin state is split in two places:
+#   - Declaration: settings.json `extraKnownMarketplaces` + `enabledPlugins` (chezmoi-owned, versioned).
+#   - Runtime: "$CLAUDE_CONFIG_DIR/plugins/" (marketplace clones, install cache, installed_plugins.json).
+#     Volatile and chezmoiignored, so it starts empty on a fresh machine.
+#
+# The CLI does not bridge the two on its own: `enabledPlugins` only toggles an already-installed
+# plugin, and `extraKnownMarketplaces` does not register the marketplace. Both gaps are upstream
+# wontfix (anthropics/claude-code#23737 closed as duplicate, #45323 closed as not planned: "Plugins
+# configured in org managed settings do not auto-install in the CLI"). Without this script, a new
+# machine gets settings.json's declaration but no plugins. Verified against claude 2.1.185: the
+# `autoInstallEnabledPlugins` and `forcedPlugins` keys those issues ask for do not exist in the binary.
+#
+# The plugin and marketplace lists below are rendered from settings.json, so the declaration has a
+# single source of truth and the run_onchange key changes only when the declaration itself changes.
+#
+# `claude plugin marketplace add` and `claude plugin install` rewrite "$CLAUDE_CONFIG_DIR/settings.json"
+# with their own serializer, which reorders the top-level keys and drops the per-hook `id`/`description`
+# annotations chezmoi ships. We snapshot the file up front and write it back on exit, so the CLI's
+# normalization never outlives this script.
+#
+# claude is mise-managed (dot_config/mise/config.toml) and installed by run_onchange_after_12, so it is
+# invoked through `mise exec` rather than PATH, for the same reason as run_onchange_after_13-setup-mcp.
+
+{{ $settings := include "dot_claude/settings.json" | fromJson }}
+if ! command -v mise &>/dev/null; then
+  echo "WARNING: mise is not installed; skipping Claude Code plugin setup."
+  exit 0
+fi
+
+if ! command -v jq &>/dev/null; then
+  echo "WARNING: jq is not installed; skipping Claude Code plugin setup."
+  exit 0
+fi
+
+# Rendered from settings.json .enabledPlugins — only the entries switched on. Each id is
+# "<plugin>@<marketplace>", which is also how the CLI addresses them.
+plugins=(
+{{- range $id, $enabled := $settings.enabledPlugins }}
+{{- if $enabled }}
+  "{{ $id }}"
+{{- end }}
+{{- end }}
+)
+
+# Rendered from settings.json .extraKnownMarketplaces as "<name>=<source>", where <source> is the
+# argument `claude plugin marketplace add` accepts (an owner/repo shorthand, or a git URL).
+declared_marketplaces=(
+{{- range $name, $marketplace := $settings.extraKnownMarketplaces }}
+  "{{ $name }}={{ if eq $marketplace.source.source "github" }}{{ $marketplace.source.repo }}{{ else }}{{ $marketplace.source.url }}{{ end }}"
+{{- end }}
+)
+
+# claude-plugins-official ships with Claude Code, so it is not an "extra" marketplace and must not be
+# declared in settings.json. It still has to be registered before any of its plugins can be installed.
+OFFICIAL_MARKETPLACE="claude-plugins-official"
+OFFICIAL_MARKETPLACE_SOURCE="anthropics/claude-plugins-official"
+
+# Each account is isolated by CLAUDE_CONFIG_DIR (see dot_config/zsh/claude.zsh: cld -> ~/.claude,
+# cld-r06 -> ~/.claude-r06). The two share one settings.json via a symlink, but each keeps its own
+# plugins/ runtime, so every account has to be reconciled separately.
+config_dirs=("$HOME/.claude" "$HOME/.claude-r06")
+
+if [ "${#plugins[@]}" -eq 0 ]; then
+  echo "==> No plugins declared in settings.json; nothing to reconcile."
+  exit 0
+fi
+
+settings_file="$HOME/.claude/settings.json"
+settings_snapshot="$(mktemp)"
+cp "$settings_file" "$settings_snapshot"
+
+restore_settings() {
+  # Undo the CLI's rewrite. Written through the existing path so the file keeps its inode.
+  cat "$settings_snapshot" >"$settings_file"
+  rm -f "$settings_snapshot"
+  # `--scope user` targets "$CLAUDE_CONFIG_DIR/settings.json", which for the work account is the
+  # symlink deployed from dot_claude-r06/symlink_settings.json.tmpl. A CLI that rewrites the file
+  # atomically (temp file + rename) would leave a regular file in the symlink's place.
+  local work_settings="$HOME/.claude-r06/settings.json"
+  if [ -e "$work_settings" ] && [ ! -L "$work_settings" ]; then
+    ln -sfn "$settings_file" "$work_settings"
+  fi
+}
+trap restore_settings EXIT
+
+# Map a marketplace name to the source argument for `claude plugin marketplace add`.
+marketplace_source() {
+  local name="$1" entry
+  for entry in ${declared_marketplaces[@]+"${declared_marketplaces[@]}"}; do
+    if [ "${entry%%=*}" = "$name" ]; then
+      printf '%s' "${entry#*=}"
+      return 0
+    fi
+  done
+  if [ "$name" = "$OFFICIAL_MARKETPLACE" ]; then
+    printf '%s' "$OFFICIAL_MARKETPLACE_SOURCE"
+    return 0
+  fi
+  return 1
+}
+
+failed=0
+for config_dir in "${config_dirs[@]}"; do
+  echo "==> Reconciling Claude Code plugins in ${config_dir}..."
+  known_marketplaces="${config_dir}/plugins/known_marketplaces.json"
+  # `plugin list --json` prints each installed plugin's "<plugin>@<marketplace>" id. It is empty (and
+  # exits non-zero) before the plugin runtime exists, which is the expected state on a first apply.
+  installed="$(CLAUDE_CONFIG_DIR="$config_dir" mise exec -- claude plugin list --json 2>/dev/null | jq -r '.[].id' || true)"
+
+  for plugin in "${plugins[@]}"; do
+    marketplace="${plugin##*@}"
+
+    if [ ! -f "$known_marketplaces" ] || ! jq -e --arg m "$marketplace" 'has($m)' "$known_marketplaces" >/dev/null 2>&1; then
+      if ! source_spec="$(marketplace_source "$marketplace")"; then
+        printf 'ERROR: plugin %s needs marketplace %s, which is neither %s nor declared in settings.json extraKnownMarketplaces\n' "$plugin" "$marketplace" "$OFFICIAL_MARKETPLACE" >&2
+        failed=1
+        continue
+      fi
+      if ! out="$(CLAUDE_CONFIG_DIR="$config_dir" mise exec -- claude plugin marketplace add "$source_spec" 2>&1)"; then
+        printf 'ERROR: failed to add marketplace %s (%s) in %s\n%s\n' "$marketplace" "$source_spec" "$config_dir" "$out" >&2
+        failed=1
+        continue
+      fi
+    fi
+
+    if printf '%s\n' "$installed" | grep -qxF "$plugin"; then
+      continue
+    fi
+    if ! out="$(CLAUDE_CONFIG_DIR="$config_dir" mise exec -- claude plugin install "$plugin" --scope user 2>&1)"; then
+      printf 'ERROR: failed to install plugin %s in %s\n%s\n' "$plugin" "$config_dir" "$out" >&2
+      failed=1
+    fi
+  done
+done
+
+if [ "$failed" -ne 0 ]; then
+  # Exit non-zero so chezmoi does not record this run as successful and retries on the next apply,
+  # instead of freezing a partially-reconciled state (same contract as run_onchange_after_13-setup-mcp).
+  echo "ERROR: one or more Claude Code plugins failed to install; chezmoi will retry on the next apply." >&2
+  exit 1
+fi
+
+echo "==> Claude Code plugins reconciled."
--- a/home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
+++ b/home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
@@ -9,14 +9,18 @@ set -euo pipefail
 #     Volatile and chezmoiignored, so it starts empty on a fresh machine.
 #
 # The CLI does not bridge the two on its own: `enabledPlugins` only toggles an already-installed
-# plugin, and `extraKnownMarketplaces` does not register the marketplace. Both gaps are upstream
-# wontfix (anthropics/claude-code#23737 closed as duplicate, #45323 closed as not planned: "Plugins
-# configured in org managed settings do not auto-install in the CLI"). Without this script, a new
-# machine gets settings.json's declaration but no plugins. Verified against claude 2.1.185: the
-# `autoInstallEnabledPlugins` and `forcedPlugins` keys those issues ask for do not exist in the binary.
+# plugin, and `extraKnownMarketplaces` does not register the marketplace -- nor is its `ref` honoured
+# unless the ref is passed on the command line. Both gaps are upstream wontfix
+# (anthropics/claude-code#23737 closed as duplicate, #45323 closed as not planned: "Plugins configured
+# in org managed settings do not auto-install in the CLI"). Without this script, a new machine gets
+# settings.json's declaration but no plugins. Verified against claude 2.1.205 (the mise-pinned version
+# in dot_config/mise/config.toml): the `autoInstallEnabledPlugins` and `forcedPlugins` keys those
+# issues ask for do not exist in the binary.
 #
-# The plugin and marketplace lists below are rendered from settings.json, so the declaration has a
-# single source of truth and the run_onchange key changes only when the declaration itself changes.
+# The declaration below is rendered from settings.json, so it has a single source of truth and the
+# run_onchange key changes only when the declaration itself changes. It is emitted as JSON inside a
+# quoted heredoc rather than as bash array literals: a value containing a quote or `$(...)` would
+# otherwise be evaluated as script source at render time, which no amount of runtime quoting fixes.
 #
 # `claude plugin marketplace add` and `claude plugin install` rewrite "$CLAUDE_CONFIG_DIR/settings.json"
 # with their own serializer, which reorders the top-level keys and drops the per-hook `id`/`description`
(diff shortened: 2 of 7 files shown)


home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+# claude is mise-managed (dot_config/mise/config.toml) and installed by run_onchange_after_12, so it is
+# invoked through `mise exec` rather than PATH, for the same reason as run_onchange_after_13-setup-mcp.
+
+{{ $settings := include "dot_claude/settings.json" | fromJson }}
+if ! command -v mise &>/dev/null; then
+  echo "WARNING: mise is not installed; skipping Claude Code plugin setup."
+  exit 0
+fi
+
+if ! command -v jq &>/dev/null; then
```
kryota-dev: [MUST] Taking the `exit 0` path when `jq` is absent marks this `run_onchange` as done, so it never retries once jq lands. On a fresh machine `jq` arrives via mise (`dot_config/mise/config.toml:26`), so like `claude` it may not be on PATH when this script runs.

The repo already knows this trap. `run_onchange_after_14-enable-clv2-observer.sh.tmpl:23` says it outright:

> exit 0: a silent skip would mark this run_onchange "done" and never retry once jq lands.

Adopt that script's `jq_cmd=(jq)` / `jq_cmd=(mise exec -- jq)` selection and exit **non-zero** when neither works. The `jq` calls further down must go through `"${jq_cmd[@]}"` too.

(The `mise` guard's `exit 0` is the existing convention from scripts 12/13/18 and can stay.)

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+  echo "WARNING: jq is not installed; skipping Claude Code plugin setup."
+  exit 0
+fi
+
+# Rendered from settings.json .enabledPlugins — only the entries switched on. Each id is
+# "<plugin>@<marketplace>", which is also how the CLI addresses them.
+plugins=(
+{{- range $id, $enabled := $settings.enabledPlugins }}
+{{- if $enabled }}
+  "{{ $id }}"
```
kryota-dev: [MUST] Values from `settings.json` are rendered directly into a double-quoted bash array literal (same on the `declared_marketplaces` line). A value containing `"`, a backtick, `$(...)`, or a newline is evaluated as **script source** after rendering. Runtime quoting like `"$source_spec"` cannot close this path - it is a different one.

Exploiting it requires write access to settings.json, so today's risk is small, but a defensive layer is missing. Emitting the declaration with `toJson` into a quoted heredoc and reading it back with `jq -r` means no value is ever evaluated by the shell.

(A quoted heredoc is not expanded, and single-line JSON cannot collide with the delimiter. Lint's `sed '/{{/d'` deletes the body without breaking the syntax.)

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+{{- end }}
+{{- end }}
+)
+
+# Rendered from settings.json .extraKnownMarketplaces as "<name>=<source>", where <source> is the
+# argument `claude plugin marketplace add` accepts (an owner/repo shorthand, or a git URL).
+declared_marketplaces=(
+{{- range $name, $marketplace := $settings.extraKnownMarketplaces }}
+  "{{ $name }}={{ if eq $marketplace.source.source "github" }}{{ $marketplace.source.repo }}{{ else }}{{ $marketplace.source.url }}{{ end }}"
```
kryota-dev: [SHOULD] `extraKnownMarketplaces[].source` has seven variants in the schema, each with a different identifying field (fetched from `json.schemastore.org/claude-code-settings.json`):

| source | required field |
|---|---|
| `url` | `url` |
| `hostPattern` | `hostPattern` |
| `github` | `repo` (optional `ref` / `path`) |
| `git` | `url` (optional `ref` / `path`) |
| `npm` | `package` |
| `file` / `directory` | `path` |

The current `github ? repo : url` dichotomy happens to be compatible with `git`, but declaring `npm`/`file`/`directory`/`hostPattern` would reference a field that does not exist. `tests/files.bats` shares the same wrong dichotomy, so a future addition would fail for an unclear reason.

Handle `github` and `git` explicitly (with `ref` - see the settings.json thread) and **fail loudly** on the rest.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+# plugins/ runtime, so every account has to be reconciled separately.
+config_dirs=("$HOME/.claude" "$HOME/.claude-r06")
+
+if [ "${#plugins[@]}" -eq 0 ]; then
+  echo "==> No plugins declared in settings.json; nothing to reconcile."
+  exit 0
+fi
+
+settings_file="$HOME/.claude/settings.json"
+settings_snapshot="$(mktemp)"
```
kryota-dev, line 85: [SHOULD] The `trap` is armed after `mktemp` and `cp`, so a failing `cp` (which `set -e` turns into an immediate exit) leaves the temp file behind in `$TMPDIR`. Unlikely - settings.json is written in the files phase first - but moving the `trap` to just after `mktemp`, before `cp`, closes it.

Note that arming it earlier means `restore_settings` can run with an empty snapshot, which would truncate settings.json; guard the restore on `[ -s "$settings_snapshot" ]`.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+  return 1
+}
+
+failed=0
+for config_dir in "${config_dirs[@]}"; do
+  echo "==> Reconciling Claude Code plugins in ${config_dir}..."
+  known_marketplaces="${config_dir}/plugins/known_marketplaces.json"
+  # `plugin list --json` prints each installed plugin's "<plugin>@<marketplace>" id. It is empty (and
+  # exits non-zero) before the plugin runtime exists, which is the expected state on a first apply.
+  installed="$(CLAUDE_CONFIG_DIR="$config_dir" mise exec -- claude plugin list --json 2>/dev/null | jq -r '.[].id' || true)"
```
kryota-dev: [SHOULD] The output of `plugin list --json` is not filtered by scope. Since the goal is `--scope user`, a plugin present only in project or local scope could mask a missing user-scoped one (each element carries a `scope` field). `select(.scope == "user")` is the robust check.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+# Plugin state is split in two places:
+#   - Declaration: settings.json `extraKnownMarketplaces` + `enabledPlugins` (chezmoi-owned, versioned).
+#   - Runtime: "$CLAUDE_CONFIG_DIR/plugins/" (marketplace clones, install cache, installed_plugins.json).
+#     Volatile and chezmoiignored, so it starts empty on a fresh machine.
+#
+# The CLI does not bridge the two on its own: `enabledPlugins` only toggles an already-installed
+# plugin, and `extraKnownMarketplaces` does not register the marketplace. Both gaps are upstream
+# wontfix (anthropics/claude-code#23737 closed as duplicate, #45323 closed as not planned: "Plugins
+# configured in org managed settings do not auto-install in the CLI"). Without this script, a new
+# machine gets settings.json's declaration but no plugins. Verified against claude 2.1.185: the
```
kryota-dev: [NITS] "Verified against claude 2.1.185" names the wrong version. 2.1.185 is only what the native installer left in `~/.local/share/claude/versions/`; chezmoi and mise install - and `mise exec` runs - **2.1.205** (`dot_config/mise/config.toml:17`).

Re-running `strings` against the 2.1.205 binary gives 32 hits for `enabledPlugins` and 18 for `extraKnownMarketplaces`, and zero for `autoInstallEnabledPlugins` and `forcedPlugins`, so **the claim does hold for 2.1.205**. Only the version needs correcting (in the docs and the PR body too).

kryota-dev (commented):
## PR #292 — integrated review

Ran cc-code-review, cc-security-review, and codex in parallel, verified each claim against a primary source, then merged the results. No prior reviews existed, so nothing was deduplicated.

### Totals

| Category | cc-code-review | cc-security-review | codex | Merged (after verification) |
|---|---|---|---|---|
| MUST | 0 | 0 | 2 | **3** |
| SHOULD | 2 | 1 (Medium) | 2 | 4 |
| NITS | 4 | 3 (Info) | 1 | 5 |
| GOOD | 5 | 6 | 4 | — |

### Security

- Overall risk: **Medium**
- Findings: 6 (Medium 1 / Low 2 / Info 3). One Medium and one Low were promoted to MUST.

The only substantive risk is that the `openai/codex-plugin-cc` marketplace is **installed unattended without a pin**. A plugin can execute arbitrary code inside a Claude Code session, and `docs/agents/skills-provenance.md` requires external code to be pinned.

### Claims that verification corrected or rejected

Nothing was dropped silently; each verdict is recorded with its evidence.

**1. cc-security-review's remedy "pin to a SHA" — corrected (not possible)**

Measured in an isolated `CLAUDE_CONFIG_DIR`:

| Method | Pinned? | Recorded in `known_marketplaces.json` |
|---|---|---|
| `marketplace add openai/codex-plugin-cc#v1.0.4` | Yes — checks out `807e03a` | `{"source":"github","repo":"...","ref":"v1.0.4"}` |
| Declaring `ref` in settings.json only | **No — HEAD is checked out** | `ref` is not even recorded |
| `#<commit-sha>` | No — `fatal: Remote branch <sha> not found` (`git clone --branch` semantics) |

So a pin is possible only against a **tag or branch**, and it does not take effect unless `<repo>#<ref>` is passed on the command line. The finding is right; its remedy is not. Replaced with declaring `ref` and rendering `<repo>#<ref>`.

**2. cc-security-review's "unverified" notes (the issues, and 2.1.185's behaviour) — verified**

- `anthropics/claude-code#23737` (duplicate) and `#45323` (not planned) both exist and match the description.
- But the comment's "verified against claude 2.1.185" has **the wrong version**. chezmoi installs, and `mise exec` runs, the mise-pinned **2.1.205**. Re-running `strings` against 2.1.205 gives 32 hits for `enabledPlugins` and 18 for `extraKnownMarketplaces`, and **zero** for `autoInstallEnabledPlugins` and `forcedPlugins`. The claim holds; only the version is wrong (NITS).

**3. cc-security-review [Info] "plugin hooks may bypass GateGuard" — kept, unverified**

Neither confirmed nor refuted against a primary source. Not refuted, so not dropped; recorded as a concern worth tracking. Not blocking for this PR.

**4. cc-security-review [Info] "`ln -sfn` TOCTOU" and "`trap` misses SIGKILL" — no action**

The reviewer itself concluded there is no impact, and a re-check agrees. `-f` forces the final state to be the intended one, and the next `chezmoi apply` restores settings.json.

**5. An error by the integrator (me)**

I initially doubted cc-code-review's claim that it had fetched the schema from schemastore and found seven source variants. The fault was mine: my `curl` lacked `-L` and grabbed the 176-byte redirect page from `json.schemastore.org`. Refetched from the raw URL, **cc-code-review was right**. `extraKnownMarketplaces[].source` has `url` / `hostPattern` / `github` / `git` / `npm` / `file` / `directory`, and `github` carries `ref` and `path` alongside `repo`.

### [MUST]

1. **`home/run_onchange_after_17-setup-claude-plugins.sh.tmpl:35`** — `exit 0` when `jq` is missing marks the `run_onchange` as done, so it never retries. `run_onchange_after_14-enable-clv2-observer.sh.tmpl:23` already carries a comment warning about exactly this, plus the `jq_cmd` pattern. (codex; CONFIRMED)
2. **`…:45,54`** — settings.json values are rendered into bash array literals, so a value containing `"`, a backtick, or `$(...)` is evaluated as script source. This is a different path from the runtime quoting of arguments. (codex MUST + cc-security-review Low, flagged independently)
3. **`home/dot_claude/settings.json:483`** — the marketplace is not pinned. Fixed per the measurements above, not per the original remedy. (cc-security-review Medium)

### [SHOULD]

4. **`…:54`** — only two of the schema's seven source variants are handled; `npm`/`file`/`directory`/`hostPattern` would reference a missing field. (cc-code-review)
5. **`…:74`** — the `trap` is armed after `mktemp` + `cp`, leaking the temp file if `cp` fails. (cc-code-review)
6. **`…:113`** — `plugin list --json` is not filtered by scope, so a project/local-scoped plugin can mask a missing user-scoped one. (codex)
7. **`tests/files.bats:472`** — the reconciler has only an existence check. (codex)

### [NITS]

8. `…:15` — "Verified against claude 2.1.185" should read 2.1.205.
9. `…:73` — `settings_file` duplicates `config_dirs[0]`.
10. `…:124` — `marketplace add` omits `--scope user` while `install` passes it.
11. `…:118` — a marketplace already registered under a different source is trusted as-is (worth a comment).
12. Script 13 runs `mkdir -p` per account; script 17 does not.

### [GOOD]

- Deriving the declaration from settings.json with `include | fromJson` removes the duplicate list, and makes the `run_onchange` key flip only when the declaration changes.
- `trap restore_settings EXIT` fires on both the success and the `exit 1` paths and restores the hook annotations. cc-code-review confirmed this on **real bash 3.2 with a fake CLI**, including regeneration of the symlink after an atomic rename replaced it.
- The bash 3.2 empty-array-under-`set -u` trap is avoided correctly, via `${arr[@]+"${arr[@]}"}` and an early return on an empty count.
- Under `pipefail`, a non-zero `claude plugin list --json` is absorbed by the trailing `|| true`.
- The added bats invariants are not vacuous: deliberately corrupting settings.json makes them fail (measured by cc-code-review).
- Exiting non-zero on failure so chezmoi retries matches script 13 and refuses to freeze a partially applied state.
- The `mktemp` snapshot is created 0600 under a per-user `$TMPDIR` on macOS (measured by cc-security-review).

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+# cld-r06 -> ~/.claude-r06). The two share one settings.json via a symlink, but each keeps its own
+# plugins/ runtime, so every account has to be reconciled separately.
+config_dirs=("$HOME/.claude" "$HOME/.claude-r06")
+
+if [ "${#plugins[@]}" -eq 0 ]; then
+  echo "==> No plugins declared in settings.json; nothing to reconcile."
+  exit 0
+fi
+
+settings_file="$HOME/.claude/settings.json"
```
kryota-dev: [NITS] `settings_file` duplicates `config_dirs[0]` as a hardcoded literal. Reordering `config_dirs` would silently break it; deriving it as `settings_file="${config_dirs[0]}/settings.json"` removes the coupling.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+  echo "==> Reconciling Claude Code plugins in ${config_dir}..."
+  known_marketplaces="${config_dir}/plugins/known_marketplaces.json"
+  # `plugin list --json` prints each installed plugin's "<plugin>@<marketplace>" id. It is empty (and
+  # exits non-zero) before the plugin runtime exists, which is the expected state on a first apply.
+  installed="$(CLAUDE_CONFIG_DIR="$config_dir" mise exec -- claude plugin list --json 2>/dev/null | jq -r '.[].id' || true)"
+
+  for plugin in "${plugins[@]}"; do
+    marketplace="${plugin##*@}"
+
+    if [ ! -f "$known_marketplaces" ] || ! jq -e --arg m "$marketplace" 'has($m)' "$known_marketplaces" >/dev/null 2>&1; then
```
kryota-dev: [NITS] If `known_marketplaces.json` already registers this marketplace name under a **different source**, only the name's presence is checked and the existing entry is trusted. The impact is slight for personal dotfiles - tampering needs local write access - but stating the assumption in a comment would help.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
```diff
@@ -0,0 +1,148 @@
...
+      if ! source_spec="$(marketplace_source "$marketplace")"; then
+        printf 'ERROR: plugin %s needs marketplace %s, which is neither %s nor declared in settings.json extraKnownMarketplaces\n' "$plugin" "$marketplace" "$OFFICIAL_MARKETPLACE" >&2
+        failed=1
+        continue
+      fi
+      if ! out="$(CLAUDE_CONFIG_DIR="$config_dir" mise exec -- claude plugin marketplace add "$source_spec" 2>&1)"; then
```
kryota-dev: [NITS] `marketplace add` relies on the default scope while `plugin install` below passes `--scope user` explicitly. Passing it here too is more robust against a future change of the default.

home/dot_claude/settings.json
```diff
@@ -473,8 +473,16 @@
...
   "enabledPlugins": {
-    "codex@openai-codex": true
+    "codex@openai-codex": true,
+    "claude-code-setup@claude-plugins-official": true
+  },
+  "extraKnownMarketplaces": {
+    "openai-codex": {
+      "source": {
+        "source": "github",
+        "repo": "openai/codex-plugin-cc"
```
kryota-dev: [MUST] This marketplace is not pinned. `chezmoi apply` runs unattended, so on a fresh machine it clones and installs whatever is at the **default branch HEAD** of `openai/codex-plugin-cc`. A plugin can execute arbitrary code through hooks, and `docs/agents/skills-provenance.md` requires external code to be pinned.

Measured in an isolated `CLAUDE_CONFIG_DIR`:

| Method | Result |
|---|---|
| `marketplace add openai/codex-plugin-cc#v1.0.4` | Checks out `807e03a`; recorded as `{"source":"github","repo":"...","ref":"v1.0.4"}` |
| Declaring `ref` in settings.json only | **Ignored** - HEAD is checked out |
| `#<commit-sha>` | `fatal: Remote branch <sha> not found` (the ref reaches `git clone --branch`, so a SHA cannot work) |

So declare a **tag** in the `ref` field the schema allows on a `github` source, and render `<repo>#<ref>` for the CLI argument - declaring it alone has no effect. The current tag is `v1.0.6`.

Since a SHA pin is impossible, the tag is mutable and this does not fully match the repo's commit-pin policy. Either track it with Renovate's `github-tags`, or record why that is accepted.

tests/files.bats
```diff
@@ -425,6 +425,54 @@ load helpers/setup
...
+  command -v jq >/dev/null 2>&1 || skip "jq unavailable"
+  # The template renders `repo` for github sources and `url` for everything else. An entry with
+  # neither would render an empty `marketplace add` argument.
+  jq -e '
+    .extraKnownMarketplaces
+    | all(.[]; if .source.source == "github" then (.source.repo | length) > 0 else (.source.url | length) > 0 end)
+  ' "$s" >/dev/null
+}
+
+@test "chezmoi source files exist: claude plugin reconciler script" {
```
kryota-dev, line 486: [SHOULD] The reconciler's only test is an existence check, so breaking the rendering, the `marketplace add` / `install` invocations, the snapshot restore, or the `.claude-r06/settings.json` symlink repair would all still pass.

A bats test using a temp `HOME` and fake `mise`/`claude` should cover at least: both accounts get what they are missing, a failure exits non-zero, and the EXIT trap restores settings.json. (cc-code-review ran an equivalent check by hand; making it a test prevents regressions.)

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. Selects a PATH `jq` or `mise exec -- jq`, probes it with `--version`, and exits non-zero when neither works. All `jq` calls now go through `"${jq_cmd[@]}"`. The `mise` guard keeps its `exit 0` as the existing convention.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. The declaration is now emitted as `dict "enabledPlugins" ... "extraKnownMarketplaces" ... | toJson` inside a quoted heredoc and read back with `jq`, so no rendered value reaches the shell as source. Restricting the embedded JSON to the two objects also keeps the `run_onchange` key from flipping on unrelated settings.json edits.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. `marketplace_source()` resolves `github` (`repo`) and `git` (`url`), appends `#<ref>` when a ref is declared, and calls jq's `error()` on any other source type so the run fails loudly. `tests/files.bats` was corrected in df53b41 to assert the same two types rather than the old `github`-or-`url` dichotomy.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. The trap is armed immediately after `mktemp`, before `cp`. As you noted, that lets `restore_settings` run with an empty snapshot, so the restore is now guarded on `[ -s "$settings_snapshot" ]` — otherwise it would have truncated settings.json.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. The pipeline is now `jq -r '.[] | select(.scope == "user") | .id'`.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a (script comment) and 5c7d4c3 (docs). The comment now cites 2.1.205 and notes it is the mise-pinned version that `mise exec` runs.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. `settings_file="${config_dirs[0]}/settings.json"`, and `restore_settings` derives the work account's path from `config_dirs[1]` for the same reason.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. The assumption is now stated in a comment above the presence check: an entry registered under a different source is trusted as-is, because rewriting it would need local write access to `known_marketplaces.json`, which already implies control of the machine.

home/run_onchange_after_17-setup-claude-plugins.sh.tmpl
kryota-dev: Fixed in e0c333a. `marketplace add` now passes `--scope user` explicitly. `mkdir -p "$config_dir"` was added too, for symmetry with script 13.

home/dot_claude/settings.json
kryota-dev: Fixed in e0c333a. `ref: "v1.0.6"` is declared on the `github` source, and the template renders `<repo>#<ref>` for the CLI argument, since the declared ref alone is ignored. df53b41 adds a test asserting every declared marketplace carries a ref, and asserting the pinned argument actually reaches `marketplace add`.

No Renovate rule yet: the tag bump is a manual edit, and the docs say so rather than claiming automation that does not exist.

One caveat worth stating: both accounts already have `openai-codex` registered from an earlier unpinned `marketplace add`, and the script skips registration when the name is present. The pin therefore takes effect on a fresh machine, or after the marketplace is removed and re-added.

tests/files.bats
kryota-dev: Fixed in df53b41. Three tests now render the template and run it against a throwaway `HOME` with fake `mise`/`claude` on PATH. The fake claude reproduces both behaviours that motivate the snapshot: it rewrites settings.json with its own serializer and replaces the file by rename.

They assert the pinned ref reaches `marketplace add`, both accounts receive both plugins, the hook annotations come back, the work account's symlink survives, a second run is a no-op, and a failed install exits non-zero.

Both assertions were mutation-checked: deleting the `trap` line strips the hook annotations and clobbers the symlink, and dropping the `#<ref>` suffix removes the pin from the logged argument — each makes the corresponding test fail.

PR #328, opened 2026-07-25
refactor(skills): vendor only the agent-browser discovery stub

kryota-dev:
## Summary

`home/dot_agents/skills/agent-browser/` carried 1,693 lines of vendored `references/` and
`templates/` alongside its discovery stub. Nothing refreshed them, nothing linked to them, and
they had already gone wrong. This removes them so the stub is the only curated artifact — the
state `docs/agents/skills-provenance.md` already describes — and makes the provenance test
enforce it.

## Root cause

The agent-browser CLI serves its own usage content:

```bash
agent-browser skills get core --full
```

That output always matches the installed binary, so the version pin lives in exactly one place —
`"npm:agent-browser"` in `home/dot_config/mise/config.toml` — and the skill content follows from
it. Upstream ships a single `SKILL.md` under `skills/agent-browser/` for that reason, and the stub
says so itself:

> The CLI serves skill content that always matches the installed version, so instructions never go
> stale. The content in this stub cannot change between releases, which is why it just points at
> `skills get core`.

The vendored `references/` and `templates/` were a frozen copy of that CLI output taken before
0.32. They sat outside the mise pin and outside any `.chezmoiexternal.toml` declaration, so no
mechanism could ever update them, and `SKILL.md` linked to neither — no reader would reach them.

They had already rotted. `references/commands.md` advertised:

```bash
agent-browser find placeholder "Search" type "query"
```

`type` is not an accepted action. The daemon validates against
`FIND_ACTIONS = ["click", "fill", "check", "hover", "text"]` (`cli/src/native/actions.rs`) and
rejects anything else with `Unknown action '<x>' for find`. Upstream corrected the line to `fill`
in 0.32.4, so the CLI-served copy is already right — only the vendored snapshot was wrong.

#145 introduced the stub-only policy but applied it to the specialized skills only, leaving the
core `references/` and `templates/` behind. The provenance test's name claimed the policy was
enforced while its assertions never checked for leftovers, so nothing caught the gap.

## What changed

| File | Change |
|---|---|
| `home/dot_agents/skills/agent-browser/references/*.md` (7), `templates/*.sh` (3) | Deleted. `SKILL.md` is now the only file in the directory. |
| `home/.chezmoiremove` | Removal targets for the deployed `references/` and `templates/` copies. Removing files from source does not delete an already-deployed copy. |
| `home/dot_agents/skills/agent-browser/SKILL.md` | Adds the `derive-client` entry upstream v0.33.0 lists between `dogfood` and `vercel-sandbox`. |
| `tests/skill_provenance.bats` | Asserts the source dir holds nothing but `SKILL.md`, and that `.chezmoiremove` carries the removal targets. Drops the `agent-browser` exclusion from the frontmatter test, whose stated premise no longer applies. |

### Also fixed: two more #145 leftovers

Writing that assertion surfaced that `.chezmoiremove` listed only three of the five specialized
skills. The history, verified against git: #145 (2026-06-12) removed all five from source before
`.chezmoiremove` existed at all; #174 (2026-06-22) re-curated electron/slack/dogfood and created
the file; #175 (2026-06-23) re-removed exactly those three and registered only them. `agentcore`
and `vercel-sandbox` were never re-curated, so no commit ever gave them a removal target, and
their copies had been sitting in `~/.agents/skills/` since 2026-04-14 — which is what the
provenance test's informational unmanaged-skill check had been reporting all along. Both are now
listed, and the test asserts all five names so a partial listing fails CI instead of staying an
informational note.

## Verification

- `make test`: exit 0, 167 assertions, 0 failures
- `chezmoi apply --dry-run`: confirms `references/`, `templates/`, `agentcore/`, and
  `vercel-sandbox/` are removed and the `derive-client` line is added. bats only checks that the
  `.chezmoiremove` lines exist, so this covers the part bats cannot.
- Each commit is green on its own: the test-strengthening commit lands after the deletions, so
  `make test` passes at every point in the history.

## Notes

`docs/agents/skills-provenance.md` already described agent-browser as a "discovery stub only", so
the policy statement itself was right and the source tree was the part that had drifted. But this
PR changes both the `.chezmoiremove` contents and the provenance assertions that three documents
quote verbatim, so those inventories are updated here as well (EN plus the `.ja.md` mirrors):

- `docs/agents/skills-provenance.md` — `SKILL.md` is the only curated file, all five CLI-served
  skills listed, both removal groups shown, and the deterministic assertion restated
- `docs/architecture/externals-and-pinning.md` — same two groups; the block claimed to be the
  "current entries" while listing 11 of 28, so it is relabelled as an excerpt
- `docs/architecture/chezmoi-engine.md` — hardcoded counts (already stale before this PR) replaced
  with a category-level description, counts deliberately omitted so it cannot drift again

`tests/docs_facts.bats` puts no `<!-- FACT -->` marker on these blocks, which is why CI did not
flag the drift.


kryota-dev (commented):
## Multi-review summary

Three reviewers ran in parallel — `cc-code-review`, `cc-security-review`, and `codex` (gpt-5.5). Every assertive claim below was re-verified against primary sources (upstream tags at `vercel-labs/agent-browser`, this repo's git history, chezmoi's reference docs, the local `.gitignore`) before being accepted or rejected.

|---|---|---|---|
| MUST | 1 | 0 | 0 |
| SHOULD | 6 | 0 | 1 |
| NITS | 3 | 0 | 0 |
| Security findings | – | 1 Medium / 3 Low / 4 Info | – |
| GOOD | 5 | 5 | 2 |

### Security

- **Overall risk: Low** — 0 Critical, 0 High, 1 Medium, 3 Low, 4 Info.
- Deletion scope was measured with `chezmoi apply --dry-run` against the real `$HOME`: exactly the four declared paths. No glob, no `..`, no absolute path. `SKILL.md` is modified rather than deleted, and the ~200 other skill directories are untouched.
- `~/.claude/skills` and `~/.codex/skills` are both symlinks to `~/.agents/skills`, so a single entry covers every harness with no mirror to miss.
- The removed `templates/*.sh` are mode 644 with no chezmoi `executable_` prefix, and grep finds no reference to them anywhere in the repo, in `~/.claude/settings*.json`, or in `~/.codex/config.toml`. Nothing executes them.
- Attack surface is unchanged by this PR: `SKILL.md` already told agents to read `agent-browser skills get core`, and nothing linked to the vendored `references/`. `skills get` reads package-bundled `skill-data/`, not the network, so the mise pin fixes its content.

### Accepted findings — will be fixed on this branch

1. **[MUST] The docs go stale with this PR.** `docs/agents/skills-provenance.md:132-138,150`, `docs/architecture/externals-and-pinning.md:191-194`, `docs/architecture/chezmoi-engine.md:157` and their `.ja.md` mirrors still quote the pre-PR `.chezmoiremove` contents and the pre-PR test assertions. Raised independently by codex and cc-code-review, and confirmed here. CI missed it because `tests/docs_facts.bats` puts no `<!-- FACT -->` marker on those blocks. **The PR body's "No documentation change" line is wrong and will be corrected** — the "discovery stub only" wording was already right, but this PR is what invalidated the surrounding inventory.

2. **[SHOULD] The new `.chezmoiremove` comment states false history.** Verified against git: `#145` (`da4d15f`) dropped all five skills from source and touched `.chezmoiremove` not at all; `#174` (`87fcc4b`) re-curated electron/slack/dogfood; `#175` (`454e327`) re-removed those three and registered exactly those three. There was never a single event where five left and only three were listed.

3. **[SHOULD] `find -mindepth 1 -not -name 'SKILL.md'` false-positives on `.DS_Store`.** Both `.DS_Store` and `._*` are gitignored (`.gitignore:2,6`), so they never reach git but `find` still sees them. Opening the directory in Finder would fail `make test` locally while CI stays green.

4. **[SHOULD] State that `.chezmoiremove` entries are permanent.** These are not one-shot cleanups — every future `chezmoi apply` recursively removes those paths.

### Rejected finding, with evidence

**cc-code-review [SHOULD]:** *"`type` is not an accepted action" contradicts `agent-browser find --help` (0.31.2), which lists `click, fill, type, hover, focus, check, uncheck`.*

**Rejected.** The claim in the commit message and the `.chezmoiremove` comment is correct; the help text was the stale part.

- `cli/src/native/actions.rs` at **v0.31.2**: `execute_subaction` has exactly five match arms — `click`, `fill`, `check`, `hover`, `text`. No `type`, no `focus`, no `uncheck`.
- `cli/src/native/actions.rs` at **v0.33.0**: `const FIND_ACTIONS: &[&str] = &["click", "fill", "check", "hover", "text"]`, plus a guard that rejects anything else with `Unknown action '<x>' for find`.

So the dispatcher never accepted `type` in either version. What 0.32.4 changed was the *advertised* list — upstream describes that release as aligning the advertised `find` actions with the set the dispatcher actually accepts. The 0.31.2 `--help` output quoted in the review is precisely that pre-fix advertisement. The reviewer marked their own confidence as `medium` on "which is right", which was the correct call.

### Surfaced but not fixed here

- **`.chezmoiremove` negation guard.** cc-code-review reports that a `!path` line silently disables a removal target (tested against chezmoi v2.70.5). chezmoi's reference documents `!` negation for `.chezmoiignore` only — it is **not** documented for `.chezmoiremove`, and this session could not independently reproduce the test. There are currently zero `!` lines. Worth a follow-up guard once the behaviour is confirmed; the guard would be safe either way, since a `!` line is either a silent disabler or meaningless.
- **Bidirectional guard: `.chezmoiremove` targets vs `.chezmoiexternal.toml` / `[ecc].skills`.** No collision today — all five names verified absent from both — but `[ecc].skills` grows through Renovate, and a future collision would have chezmoi fighting itself over one destination.
- **`SKILL.md`'s specialized-skill list has no sync mechanism.** It is verbatim upstream v0.33.0, but nothing re-syncs it when the mise pin moves. Same rot class as the `references/` this PR deletes.
- **`allowed-tools` includes `Bash(npx agent-browser:*)`.** `npx` resolves outside the mise pin. It is verbatim upstream, so removing it would deviate from the mirror-upstream policy.
- **`find-skills` / `patch-remote-control` runtime leftovers.** No `git log --diff-filter=A` trace, so they were never in source — a different class from the `#145` leftovers this PR cleans.
- **Upstream `scripts/postinstall.js`** downloads a native binary without checksum verification and follows redirects without host validation. Upstream concern, unrelated to this PR.

### GOOD

- `SKILL.md` is identical to upstream v0.33.0 modulo line wrapping (word-stream diff against the v0.33.0 tag). No repo-local instruction text has been injected into an agent-facing file.
- Deleting the frozen copy is a net security improvement: the CLI-served `skill-data/core/references/` carries `trust-boundaries.md` — page content is data not instructions, indirect prompt-injection entry points, never surface cookies or tokens to the model — which the deleted pre-0.32 snapshot predates.
- The test changes only tighten; no assertion was relaxed. `grep -qFx` is fixed-string and whole-line, so a skill name containing regex metacharacters cannot mismatch.
- Commit ordering keeps `make test` green at every point in history, verified by re-running the old test file against the post-deletion tree.
- The removed frontmatter-test exclusion carried a comment claiming it mirrored an exclusion in the non-empty test — no such exclusion existed. Deleting the stale comment along with the code was right.

home/.chezmoiremove
```diff
@@ -15,9 +15,21 @@
 # `agent-browser skills get <name>` (version-matched, never stale), so vendoring the
 # static SKILL.md is redundant. Only the discovery stub (dot_agents/skills/agent-browser)
 # stays managed; these previously-deployed copies are removed.
+# All five skills left the source tree together, but only the first three were listed
```
kryota-dev: [SHOULD] This history is wrong, and `.chezmoiremove` comments are permanent context for whoever reads this next. Verified against git:

| commit | date | what actually happened |
|---|---|---|
| `d5cde48` | 2026-04-14 | all five added to source |
| `da4d15f` (#145) | 2026-06-12 | all five removed from source; `.chezmoiremove` untouched |
| `87fcc4b` (#174) | 2026-06-22 | electron/slack/dogfood re-curated |
| `454e327` (#175) | 2026-06-23 | those three re-removed, and registered here for the first time |

There was never a single event where five left and three were listed — it is two separate events composed together.

Suggested replacement:

```
# #145 dropped all five from source without registering anything here; #175 re-removed
# electron/slack/dogfood and registered only those three, so agentcore and
# vercel-sandbox lingered as deployed copies.
```

home/.chezmoiremove
```diff
@@ -15,9 +15,21 @@
...
+# here at the time, so agentcore and vercel-sandbox lingered as unmanaged runtime dirs.
 .agents/skills/electron
 .agents/skills/slack
 .agents/skills/dogfood
+.agents/skills/agentcore
```
kryota-dev, line 35: [SHOULD] Worth stating that these entries are permanent rather than one-shot cleanups. Every future `chezmoi apply` recursively removes these paths, so anything a user or another tool later places at `~/.agents/skills/{agentcore,vercel-sandbox}` or `~/.agents/skills/agent-browser/{references,templates}` disappears without a prompt.

`~/.claude/skills` and `~/.codex/skills` are both symlinks to `~/.agents/skills` (verified), so one entry covers every harness — which is good, but it also means a skill created through either harness lands on the same removal path.

No behaviour change needed: the paths are correct, and `chezmoi apply --dry-run` against the real `$HOME` removes exactly these four and nothing else. This is only about making the destructive contract explicit for the next reader.

tests/skill_provenance.bats
```diff
@@ -51,6 +51,28 @@ _skill_is_external() {
       false
     }
   done
+
+  # The source dir itself must contain nothing but SKILL.md: no frozen references/
+  # or templates/ copies left behind (they'd go stale and contradict the stub-only
+  # policy, e.g. references/commands.md documenting a `find` action the daemon
+  # never supported).
+  local leftover
+  leftover="$(find "${HOME_DIR}/dot_agents/skills/agent-browser" -mindepth 1 -not -name 'SKILL.md')"
```
kryota-dev: [SHOULD] This false-positives on `.DS_Store`. Both `.DS_Store` and `._*` are gitignored (`.gitignore:2,6`), so they never reach git but `find` still sees them — opening this directory in Finder makes `make test` fail locally while CI stays green.

The policy is really "nothing but `SKILL.md` is *committed*", so listing tracked files matches the intent more closely and is robust against any untracked cruft:

```bash
local leftover
leftover="$(git -C "${REPO_ROOT}" ls-files 'home/dot_agents/skills/agent-browser/' \
  | grep -vx 'home/dot_agents/skills/agent-browser/SKILL.md' || true)"
```

The `|| true` is required — `grep -v` exits 1 when it filters everything out, which would abort the test. `REPO_ROOT` is already defined in `tests/helpers/setup.bash:4`.

Minimal alternative if you prefer to keep `find`: add `-not -name '.DS_Store' -not -name '._*'`.

The basename-matching logic itself is sound and has no hole — a nested `references/SKILL.md` still surfaces, because its parent directory `references` is reported as leftover.

home/.chezmoiremove
kryota-dev: **Fixed ( 6d7ec91 )**

Verified against git and rewrote the note. It turned out to be wrong in a further way than the review said: `.chezmoiremove` did not merely omit these two — the file **did not exist at all** at #145 (`git log --diff-filter=A` puts its creation in #174, ten days later). So #145 could not have registered anything, #175 registered exactly the three it re-removed, and `agentcore` / `vercel-sandbox` were never re-curated and therefore never got a removal target from any commit.

home/.chezmoiremove
kryota-dev: **Fixed ( 6d7ec91 )**

Added the note at file scope rather than inside this block, since the standing-declaration behaviour is a property of `.chezmoiremove` itself and applies to every entry, not just the agent-browser ones. It records that each future `chezmoi apply` re-deletes these destinations without warning, and that `~/.claude/skills` and `~/.codex/skills` are both symlinks to `~/.agents/skills` — so one entry covers both harnesses, and a skill authored from either lands on the same removal path.

tests/skill_provenance.bats
kryota-dev: **Fixed ( 6d7ec91 )**

Switched to `git ls-files` with the `|| true` guard, and rewrote the comment to say the check is about committed content rather than what happens to be on disk.

Demonstrated rather than assumed: `touch home/dot_agents/skills/agent-browser/.DS_Store`, then `bats tests/skill_provenance.bats` — all 10 tests still pass, where the `find`-based check would have reported the stray file as leftover content.

PR #335, opened 2026-07-25
feat(claude): Opus 5 pins, per-agent effort tiers, and the model-fitness gate

kryota-dev:
## Summary

Third of four PRs implementing #330, per the finalized PRD
(`.claude/prds/330-model-tier-codex-permissions.prd.md`, merged in PR-2).

- **Session pins** (AC-013/014): `claude-opus-5[1m]` and `effortLevel: "xhigh"`. The repo
  was a generation behind, and the effort setting existed only on the deployed side — so
  every `chezmoi apply` silently reset it.
- **Agent tier pins** (AC-015/016/017/018): every finding-generating agent now pins both
  `model` and `effort`; nothing is left on `inherit`. Adds `adversarial-verifier`
  (sonnet @ xhigh) — the Agent tool cannot pass `effort` per call, so a refuter at that
  tier has to exist as its own definition.
- **`model-fitness-check`** (AC-022/023): a new curated skill holding the §4 contract as
  its single source of truth. Floor rows block with switch / continue-with-a-note /
  abort; the trivial-small row is a non-blocking cost hint; capability ordering is
  monotonic, so a Fable session is never asked to switch down. The three orchestration
  skills call it instead of restating values — restating is exactly how the Codex pin
  drifted across three copies.
- **Delegation rewires** (AC-009/010/011/012/019/020/021/030): CI log triage to a Haiku
  worker escalating to Sonnet (summary diagnosis only, no raw logs in the parent
  context); large-tier adversarial verification to three `adversarial-verifier` agents
  with distinct framings; `sdd` Explore pinned by task shape (haiku for verbatim
  retrieval, sonnet for pattern recognition); `multi-review` fact-checking fanned out
  per finding with synthesis kept in the parent; the Codex leg switched to
  `codex exec ... review --base`; Codex worker routes for small-tier implementation and
  CI fixes; `codex:codex-rescue` excluded from skill orchestration.
- **Machine verification** (AC-025): a `FACT:claude-model-pin` marker plus a
  string-comparison bats assertion (the existing FACT extractor is digit-only), an
  `.effortLevel` assertion, and a sweep asserting every agent definition pins both axes.

## Behavioral change to call out

Moving `cc-code-review`, `cc-security-review`, and `renovate-analyzer` off `model: inherit`
**also affects standalone use of those agents** — invoked outside `multi-review` they now
run at sonnet with their pinned effort rather than inheriting the caller's session model.
`multi-review` documents the inverse opt-in: the caller raises the model for
[...]

PR #340, opened 2026-07-25
feat(ntfy): replace desktop-notify with a self-hosted ntfy channel over Tailscale

kryota-dev:
closes #337

Replaces the ECC `stop:desktop-notify` hook with a self-hosted ntfy channel: the official
ntfy container on this Mac, published to the tailnet by `tailscale serve`, fed by a
repo-owned hook that emits attributable, priority-tiered notifications into a 7-day
server-side history.

The channel **ships disabled** (`[ntfy].enabled = false`). It needs a one-time manual
bootstrap that chezmoi cannot perform, and until then the hooks fall back to a local
desktop notification — which is exactly what the hook they replace did.

## Corrections to the issue's stated findings

Three of the issue's premises did not survive verification. They are recorded here because
two of them changed the design.

1. **`brew "ntfy"` does not provide a server on macOS.** Upstream: *"Only the ntfy CLI is
   supported on macOS. ntfy server is currently not supported."* homebrew-core builds the
   macOS formula with `-tags noserver`; the installed binary lists only `publish` and
   `subscribe`, and `ntfy serve --help` returns *"No help topic for 'serve'"*. The issue's
   claim that the server binary was already provisioned is false. Resolved by running the
   official container under Docker Desktop (already Brewfile-managed), chosen over a
   source build and over relocating to the Linux homelab.
2. **The `Notification` hook does support matchers**, contrary to one line of the docs;
   the reference also states *"Matches on notification type. Omit the matcher to run hooks
   for all notification types"* and enumerates the eight values. This implementation omits
   the matcher deliberately and keeps the mapping in one place.
3. **`cache-duration` defaults to 12h**, not indefinitely, so the retention needed to be
   raised explicitly for the "review it later" requirement to hold.

## Design

- **Ingress.** The container publishes on `127.0.0.1` only; `tailscale serve --bg
  --https=443` fronts it with the node's tailnet certificate. That gives no LAN exposure,
  no certificate management, a stable MagicDNS name instead of an address literal, and —
  because the hostname lives in 1Password — nothing machine-identifying in a public repo.
- **No LaunchAgent.** `restart: unless-stopped` plus Docker Desktop's start-at-login
  already supervise the container; a second supervisor would race it.
- **The enable flag is load-bearing.** Ignoring a target also stops chezmoi *evaluating*
  its template (measured against a deliberately-failing template: ignored applies cleanly,
  un-ignored aborts the apply). So `onepasswordRead` is never reached on a machine or CI
  runner without the vault item, `chezmoi apply` still converges, and neither CI job needs
  a new entry in its "Exclude CI-incompatible files" step.
- **One topic, priority tiers.** Permission prompts / agent-needs-input / elicitation
  dialogs / idle prompts at 4, agent completion and turn end at 3, `StopFailure` at 5,
  lifecycle bookkeeping at 1. ntfy's per-priority channels already give per-device sound
  and DND control, so a topic per event type would have meant eight subscriptions and
  eight ACL entries for no gain.
- **Attribution.** Title `<repo>/<branch> · <account>`, plus tags for event, repo, branch,
  account and session prefix. The repository name comes from git's **common dir**, not
  `--show-toplevel`: worktrees here are named after their branch, so the toplevel basename
  would report the branch as the repository.
- **Fail-soft everywhere.** The hook is `async` with a 3s curl bound and always exits 0.
  Any failure falls back to a local notification, capped at priority ≥ 3 — min-priority
  lifecycle events exist for the history, and popping them locally would have made
  retiring a Stop-only hook a net *increase* in notification volume.
- **ECC retired by flag, not by edit.** Its runtime is a SHA-pinned external re-extracted
  on refresh, so editing it in place would be undone. `ECC_DISABLED_HOOKS` is the only
  durable control point; the settings.json block stays wired (same treatment as #280).

## Deliberate deviation from a repo convention

`run_onchange_after_31-setup-ntfy` **warns and exits 0** when the Docker daemon is
unreachable, instead of hard-failing so chezmoi retries. Docker Desktop's running state is
user-controlled and outside chezmoi's convergence model; hard-failing would break every
unrelated `chezmoi apply` whenever Docker happened to be closed. A missing compose file —
something chezmoi *does* control — still fails loudly.

## Bugs found and fixed during implementation

Found by smoke-testing and by review, all fixed in this PR:

- The worktree repo-name bug above (found by driving the hook with a real payload).
- The summary shipped Markdown markers (`## Heading`) to the lock screen; leading
  structure is now stripped before the emptiness test, so rules, fences and empty bullets
  fall through to the next line.
- A `{{ fail }}` written inside a `.chezmoiignore` **comment** was executed — that file is
  itself a template — which broke the apply. The first verification of the ignore gate also
  captured the exit code incorrectly and missed it; both were redone.
- **The embedded hashes were blind to an image bump.** `include` returns raw template
  source, so it sees the `.ntfy.image` placeholder rather than its value: bumping the pin
  left both hashes byte-identical and the reconcile script would never re-run, leaving the
  container on the old image indefinitely. Reproduced directly (`chezmoi execute-template`
  before/after a bump produced the same hash while the rendered compose changed), then
  fixed by embedding the resolved tag.
- **The image pin had no Renovate custom manager.** Every other pin in
  `.chezmoidata.toml` has an explicit `customManagers` entry; an inline `# renovate:`
  comment is honoured by specific built-in managers, not for an arbitrary TOML file, so
  the pin would never have received a bump PR. Verified against `renovate.json5` and fixed.

## Review findings addressed

- Local fallback now covers non-macOS via `notify-send` (the ECC hook covered WSL too).
  WSL toast support is **not** carried over and is documented as a known gap.
- Byte-manipulation pipelines run under `LC_ALL=C`; BSD `tr`/`sed` abort on input invalid
  in the caller's locale, which under `set -eu` would silently truncate a tag. Locked in
  with a multibyte-branch-name test.
- Load-bearing numbers (port, retention, summary length) now carry `<!-- FACT -->` markers
  checked against their sources by `docs_facts.bats`.
- The leak grep now also scans top-level markdown, and matches any `<name>.ts.net` rather
  than only the default MagicDNS shape.
- Residual risks are documented rather than left implied: assistant text is published and
  retained without secret scanning; loopback binding is not local isolation, and
  `behind-proxy` lets a local caller spoof its visitor IP, so the deny-all ACL is the real
  control; the web app stays reachable on the tailnet.

### Findings deliberately not acted on

- **Token in the hook's `curl` argv.** Accepted, not fixed: reading our argv requires the
  same UID that can already read the `0600` env file directly, so no privilege boundary is
  crossed and no additional party gains access. Fixing it would trade a testable code path
  for process substitution plus curl-config parsing with its own failure modes. Documented.
- **Secret-pattern redaction of the notification body.** Out of the approved scope, and
  partial pattern matching would suggest a guarantee it cannot make. The residual risk is
  documented instead, which was the reviewer's stated minimum.
- **`web-root: disable`.** The web app is a legitimate subscriber UI on the tailnet and
  reading messages through it still requires the token.

## Open questions from the PRD

1. **`tailscale serve` operator permission — resolved.** Verified on this machine: the
   exact command the setup script runs succeeds without `sudo` and without any operator
   change (exit 0), and the string the idempotency check greps for appears in real
   `tailscale serve status` output. State was restored afterwards. The script still
   fails soft with the `sudo tailscale set --operator` remedy for other machines.
2. **ntfy tag charset — closed by design.** Tag components are sanitised to
   `[A-Za-z0-9_-]`, so no assumption about separators is relied on.
3. **Rate limits — deferred to bootstrap.** Cannot be measured before the server runs; the
   mitigation (a `visitor-request-limit-*` bump) is a config change, not a design change.

## Verification

- `make lint` and `make test` pass: **217 tests, 0 failures** (34 new: 19 behavioural, 15
  static/doc-facts).
- `chezmoi apply --dry-run` verified on both paths — disabled (no ntfy targets, `.config/ntfy`
  in the ignored set, 1Password never reached) and enabled with stubbed secrets (`server.yml`
  and `notify-env` at `0600`, compose at `0644`, hook at `0755`, setup script renders empty
  when disabled and with both hashes when enabled).
- Rebased onto `main` after #335 landed; full suite re-run green on the rebased tree.

## Not included

`SubagentStop` / `TaskCompleted` wiring, phone-app subscription automation, ntfy
user/token bootstrap automation, repo-radar (#257) integration, the `notify` zsh alias,
ntfy Web Push / attachments, and Docker Desktop's start-at-login preference.

## Bootstrap required before this does anything

Create the `Dotfiles - ntfy` 1Password item, flip `[ntfy].enabled`, create the ntfy user,
ACL and token inside the container, then subscribe from each device. Full procedure in
`docs/architecture/notifications.md`.

---

## Post-review update

A four-reviewer round (code, security, architecture, Codex) ran after the first push. It
found six MUST-level defects, all reproduced first-hand before being fixed — the full
record, including the findings that were deliberately **not** acted on and why, is in the
review comment on this PR. The ones that changed behaviour:

- **A missing `server.yml` was a fail-open.** Docker's short bind syntax creates a missing
  source as an empty *directory*, so ntfy would have started on its built-in defaults
  (`auth-default-access: read-write`, no auth-file) while `tailscale serve` kept publishing
  it — an anonymous read-write server on the tailnet. Now long syntax with
  `create_host_path: false`, plus a config check in the reconcile script.
- **`squote` does not escape an embedded single quote**, and the env file it renders is
  *sourced*. A vault value containing `'` became executable code (measured). Guarded
  fail-closed at render time.
- **`docker compose up -d` does not recreate on a bind-mount content change**, so config
  edits would have rendered to disk and never taken effect. Now `--force-recreate`.
- **`StopFailure` carries `error`, `SessionEnd` carries `reason`** — neither has `message`.
  Both were publishing a literal `(no message text)`, including the only priority-5 event
  this channel has.
- **The disabled state raised local notification volume roughly sevenfold.** Since the
  channel ships disabled, that is the state every machine lands in. Falling back for every
  priority-3-and-up event made removing a `Stop`-only hook a net *increase* — the opposite
  of the invariant this design set itself, and the docs claimed the opposite of what the
  code did. Unbootstrapped now falls back for turn-end events only.
- **A Funnel guard added mid-review self-tripped.** `tailscale funnel status` and `serve
  status` print the same config, distinguished only by an annotation, so matching the
  target there fired on the state the script itself creates and would have failed every
  later `chezmoi apply`. Replaced with a structural `AllowFunnel` check.

Also: `osascript` moved to argv passing (the pattern `morning-radar.sh` already used here,
which the first implementation missed and re-invented worse), the publish checks the HTTP
status explicitly instead of trusting `--fail` with 3xx, the image pin joined the
`automerge: false` carve-out for third-party executable code, and the leak-detection test
was rewritten around `git ls-files` with a positive control after it turned out to pass
vacuously if the scan broke.


home/dot_config/ntfy/compose.yaml.tmpl
```diff
@@ -0,0 +1,64 @@
...
+          create_host_path: false
+      # Message cache (cache.db) and auth database (user.db). Kept on the host so history
+      # and credentials survive image upgrades and `docker compose down`. The setup script
+      # creates this directory (0700) before the container starts.
+      - type: bind
+        source: "{{ .chezmoi.homeDir }}/.local/state/ntfy"
+        target: /var/lib/ntfy
+        bind:
+          create_host_path: false
+    # Baseline hardening. Deliberately NOT dropping all capabilities or setting `user:`:
```
kryota-dev, line 52: [SHOULD] 意図的に見送った項目の記録。`cap_drop: ALL` と `user:` はコンテナのハードニングとして正しい方向ですが、ntfy はコンテナ内で :80 を bind するため `CAP_NET_BIND_SERVICE` を落とすと起動しません。内部ポートを 1024 以上へ移す変更とセットでしか成立せず、Docker daemon が起動しておらず vault アイテムも未作成の現状では実際に起動して確かめられません。**起動検証できない状態で分割適用するのは避け**、`no-new-privileges` と資源制限のみ先に入れています。ブートストラップ時にコンテナが起動したら、ポート変更 + `cap_drop` + `user:` を一括で扱うのが筋です。

home/dot_claude/executable_ntfy-notify.sh
```diff
@@ -0,0 +1,303 @@
...
+    # Explicit status check rather than --fail: --fail only treats >= 400 as an error, so a
+    # 3xx would be reported as success and the publish would be silently lost with no
+    # fallback. --proto '=https' refuses a plaintext base URL (the token travels in a
+    # header), --noproxy '*' keeps the body and token away from any proxy the hook's
+    # environment happens to define, and `--` stops a URL beginning with a dash being read
+    # as an option. A connection failure yields 000 and falls through to the fallback.
+    http_code="$(printf '%s' "$body" | curl --silent \
+      --proto '=https' --noproxy '*' \
+      --max-time "$CURL_MAX_SECONDS" \
+      --header "Authorization: Bearer ${NTFY_TOKEN}" \
```
kryota-dev, line 273: [NITS] トークンが `curl` の argv に載る件は **受容**（修正しない）です。argv を読めるプロセスは同一 UID で動いており、その時点で 0600 の `notify-env` を直接読めます。**権限境界を跨がないため、トークンを取得できる主体は増えません**。

`--config /dev/fd/3` でプロセス置換経由にすれば argv から消せる、という低コストな代替があることは認識しています。採らなかったのは、curl の config 解析という独自の失敗経路が増え、かつ「トークンが body ではなくヘッダにのみ乗る」ことを検証している bats アサーションが弱くなるためです。`docs/architecture/notifications.md` の残存リスク節に受容根拠として記載しています。

home/run_onchange_after_31-setup-ntfy.sh.tmpl
```diff
@@ -0,0 +1,135 @@
...
+# over the RAW template source via `include`, which does not evaluate the template — so
+# rendering this script never reaches 1Password.
+#
+# That also means the hashes are blind to .chezmoidata.toml: `include` sees the literal
+# `.ntfy.image` placeholder, not its value, so a Renovate bump of the image pin would leave
+# both hashes byte-identical and this script would never re-run — the container would stay
+# on the old image forever. The evaluated image line below closes that hole by putting the
+# resolved value into the script body itself.
+# compose hash: {{ include "dot_config/ntfy/compose.yaml.tmpl" | sha256sum }}
+# server config hash: {{ include "dot_config/ntfy/private_server.yml.tmpl" | sha256sum }}
```
kryota-dev, line 19: [SHOULD] 既知の穴として記録。`include` は生テンプレートを返すため、このハッシュは **1Password の値の変更に対しても盲目**です。vault の `base_url` / `topic` を変えると `~/.config/ntfy/server.yml` は書き換わりますが、スクリプト本文は 1 バイトも変わらないので再実行されず、`--force-recreate` にも到達しません。

`.chezmoitemplates/` へ本体を移して `includeTemplate` でハッシュを取れば塞げます（このリポジトリの `codex-model-pin.toml` と同じ形）。採らなかったのは、`base_url` の変更がマシン改名時に限られる低頻度事象であり、リファクタの範囲に対して見合わないためです。代わりに `docs/architecture/notifications.md` の残存リスク節と Operations に手動収束手順を記載しました。

home/run_once_after_11-validate-1password.sh.tmpl
```diff
@@ -69,4 +69,27 @@ if command -v python3 &>/dev/null; then
   fi
 fi
 
+{{ if .ntfy.enabled -}}
+# Self-hosted ntfy channel (#337). Deliberately validated OUTSIDE the ITEMS array above:
+# _onepassword_item_list() in tests/helpers/setup.bash parses that array to drive the
+# FACT:onepassword-vault-item-count marker in the docs, and this item is only required once
+# [ntfy].enabled is true. Its own guarded block keeps the documented count honest for the
+# default (disabled) configuration while still failing loudly for an enabled one.
+echo "==> Verifying the ntfy channel's 1Password item..."
```
kryota-dev, line 78: [NITS] このブロックの親切なエラーメッセージは、実際には**表示されない可能性が高い**という指摘がありました。`run_once_after_11` は chezmoi の二相モデル上ファイル書き出しの**後**に走るため、`[ntfy].enabled = true` で vault アイテムが無い場合、`.config/ntfy/server.yml` のレンダリング時点で `onepasswordRead` が生のエラーを出して apply を止めます。

既存の 4 アイテム（AWS / Exa / Firecrawl / Redact Patterns）もまったく同じ構造なので、**既存パターンの踏襲として受容**しました。before フェーズへ移すのは全アイテム共通の別課題です。

kryota-dev (commented):
## Multi-Review 統合レビュー結果 (PR #340)

4 レビュアー（cc-code-review / cc-security-review / architecture-reviewer / codex）を並列実行。**MUST 6 件・SHOULD 相当 9 件を修正し、`848be90` までで push 済み**です。以下は記録であり、指摘のほとんどは既にこの PR 内で解消されています。

### 総合評価

| カテゴリ | cc-code-review | cc-security-review | architecture-reviewer | codex |
|---|---|---|---|---|
| MUST | 3 | 1 (High) | 1 | 1 |
| SHOULD | 8 | 3 (Medium) | 3 | 4 |
| NITS | 11 | 6 (Low/Info) | 2 | 1 |
| GOOD | 8 | 11 | 4 | 1 |

セキュリティ総合リスク: **Medium**（High 1 件を修正済み、残存は Low 以下）

### 修正した MUST（すべて一次検証で再現してから修正）

| # | 指摘 | 検証方法 | 修正 |
|---|---|---|---|
| 1 | compose の短縮 bind 構文は source 不在時に空ディレクトリを自動生成する。`server.yml` が欠けると ntfy は組み込みデフォルト（`auth-default-access: read-write`・auth-file なし）で起動し、`tailscale serve` がそれを tailnet に公開し続ける＝**匿名 read-write のフェイルオープン** | ntfy 配布 `server.yml` の既定値を一次ソースで確認 | long syntax + `create_host_path: false`、スクリプト側にも `server.yml` ガードを追加 |
| 2 | `squote` は埋め込みシングルクォートをエスケープしない。source される env ファイルなので、vault 値が**任意コマンド実行**に昇格する | `chezmoi execute-template` で `a'; touch …; 'b` が実行可能な行になることを再現 | レンダリング時に `fail` で落とす fail-closed ガード |
| 3 | 埋め込みハッシュが image bump を検知しない。`include` は生テンプレートを返すため、Renovate が pin を上げてもハッシュが不変で `run_onchange` が発火せず、コンテナが旧イメージのまま残る | bump 前後で `chezmoi execute-template` のハッシュ行が同一なのに compose の描画結果は変わることを再現 | 解決済み `.ntfy.image` をスクリプト本文に埋め込み |
| 4 | image pin に Renovate custom manager が無く、そもそも更新 PR が来ない | `renovate.json5` の既存 pin がすべて `customManagers` を持つことを確認 | customManager を追加。さらに `automerge: false` の carve-out へ（ECC と同じ扱い） |
| 5 | `docker compose up -d` は **service definition** が変わったときのみコンテナを再作成する。bind mount の**中身**の変更は不可視で、ntfy は起動時に一度だけ設定を読むため、保持期間や ACL の変更が永久に反映されない | Compose の再作成条件 | `--force-recreate` を追加 |
| 6 | `StopFailure` は `error`、`SessionEnd` は `reason` を運び、`message` を持たない。両方とも本文が `(no message text)` になっていた＝**唯一の priority 5 イベントが中身ゼロ** | インストール済み Claude Code バイナリからスキーマ抽出、実 payload で再現 | `.error` / `.reason` をフィールド順に追加 |

### 自分がこのレビュー中に作り込んだ回帰（検出・修正済み）

- **Funnel ガードの自己誤検知**: `tailscale funnel status` と `serve status` は**同一の設定**を出力し、`(tailnet only)` / `(Funnel on)` の注記でしか区別できません。ターゲット文字列を grep するガードは、**自分のスクリプトが作った状態で必ずトリップし、以後の `chezmoi apply` を毎回 exit 1 で壊します**。実機で `TRIPS` を再現し、`AllowFunnel` を見る構造的判定へ置換しました（`AllowFunnel` は Funnel 無効時 `null` であることを実測）。

### 設計上の欠陥（コード修正で対応）

**無効状態でのローカル通知量が ECC 比 7 倍に増えていた。** この機能は無効で配布されるため、マージ直後の全マシンがその状態です。フォールバックが priority 3 以上すべてで発火していたため、`Stop` のみだった ECC フックの廃止が通知量の**増加**になっていました（PRD 自身が掲げた不変条件の逆）。ドキュメントも「置き換え前と同じ挙動」と**事実と異なる主張**をしていました。

→ 未ブートストラップ時はターン終了イベント（`Stop` / `StopFailure`）のみにフォールバックを限定し、設定済みなら従来どおり全 attention イベントで発火。集計不変条件を bats で固定し、docs も訂正しました。

### その他の修正

- `osascript` を argv 渡し（`on run argv`）へ。**同一ディレクトリの `morning-radar.sh` に既にこのパターンがあった**のに再発見できず、非可逆な劣化版（バックスラッシュ除去・引用符置換）を実装していました
- `curl --fail` は 3xx を成功扱いにするため、明示的な 2xx 判定へ。加えて `--proto '=https'` / `--noproxy '*'` / `--` / https スキーマ検証
- `notify-send` の body と title を XML エスケープ（body-markup を解釈するデーモン向け。ブランチ名に `<` は git 上有効）
- `LC_ALL=C` をバイト処理に付与、`HOME` 未設定でも exit 0
- `.chezmoiignore` の ntfy ゲートに OS 条件を追加（Linux では検証も再構成も走らないため）
- 漏洩検知テストを `git ls-files` 全走査 + `-i` + `tskey-` + **positive control** へ（旧実装は走査が全滅しても緑になる vacuous pass だった）
- ループバックテストを「全ポートエントリが loopback」へ（旧実装は追加の全インタフェース公開を見逃した。かつ BSD sed が `\s` を解さず範囲が空だった）

### 意図的に見送った指摘（理由付き）

| 指摘 | 判断 |
|---|---|
| `curl --header` でトークンが argv に露出 | **受容**。argv を読めるプロセスは同一 UID で動いており、その時点で 0600 の env ファイルを直接読めます。権限境界を跨がず、取得できる主体は増えません |
| 通知本文の secret redaction | **不採用**。承認済み PRD のスコープ外であり、部分的なパターン照合は保証できないものを保証したかのように見せます。残存リスクとして明記 |
| `web-root: disable` | **不採用**。web UI は tailnet 上で履歴を閲覧する正規の手段で、閲覧には結局トークンが要ります |
| `cap_drop: ALL` / `user:` | **見送り**。ntfy はコンテナ内で :80 を bind するため `CAP_NET_BIND_SERVICE` の除去は起動を壊します。ポート変更と併せて一括で扱うべきで、コンテナを起動できない状態での分割変更は避けました。`no-new-privileges` と資源制限は追加済み |
| `pinDigests` | **見送り**。compose と customManager の両方を digest 対応にする必要があり、automerge carve-out で当面のリスクは解消されます |
| vault 値変更時にサーバが再起動しない | **文書化**。`includeTemplate` 化で塞げますが、`base_url` 変更はマシン改名時のみで頻度が低く、手動収束手順を docs に記載しました |
| `run_once_after_11` は書き出し**後**に走るため、親切なエラーが表示される前に `onepasswordRead` が apply を止める | **既存パターン踏襲として受容**。既存の 4 アイテムも同じ構造です |

### 検証

- `make lint` + `make test`: **217 tests / 0 failures**（新規 34: 挙動 19・静的 15）
- `chezmoi apply --dry-run` を無効・有効の両経路で実行（無効: ntfy ターゲット 0・ignored・スクリプト 0 バイト・1Password 未到達 / 有効: 4 ターゲット・0600 レンダリング）
- `fail` ガードが実際に apply を中断することを確認
- `tailscale serve` の適用と原状復帰を実機で実施（PRD Open Question 1 の解決）

### 補足

`Setup Validation` ワークフローは `disabled_manually` のため、実 `chezmoi apply` の CI 検証はこの PR では走っていません。代替として上記のローカル実機検証を行っています。

PR #341, opened 2026-07-25
fix(clv2): move the homunculus dir out of the Claude config dir so instincts can be written

kryota-dev:
## Why

CLV2's Haiku observer had never produced a single instinct. `instincts/personal/` and
`instincts/inherited/` were empty across every project, so the ECC SessionStart instinct
injection block was always empty, and every analysis cycle burned Haiku cost for zero output.

#336 filed this as an active-hours plus `--max-turns` problem. Those are real, but they are
symptoms. The root cause, confirmed by direct measurement, is that `CLV2_HOMUNCULUS_DIR`
pointed inside the Claude Code **config dir**:

1. `_claude_with_home` set `CLV2_HOMUNCULUS_DIR="$home_dir/ecc-homunculus"`, i.e.
   `~/.claude/ecc-homunculus` and `~/.claude-r06/ecc-homunculus`.
2. Claude Code classifies every path under the config dir as a *sensitive file* and requires
   interactive approval before a write. Measured verbatim:

   ```
   Claude requested permissions to edit
   /Users/<user>/.claude/ecc-homunculus/projects/<id>/instincts/personal/zz-probe.md
   which is a sensitive file.
   ```

3. The CLV2 analysis pass is headless — `claude --model haiku --print --allowedTools "Read,Write"`
   (`continuous-learning-v2/agents/observer-loop.sh`) — so nobody can approve it.
4. Haiku therefore spent its turn budget negotiating permission and ended at
   `Error: Reached max turns (20)` (37 occurrences in one project's observer log). Instinct
   count stayed at exactly zero.

Two alternatives were tested and rejected against primary evidence:

| Option | Measured result |
|---|---|
| Write outside the config dir (`~/.local/share/...`) while headless | **Succeeds** — adopted |
| `permissions.allow` with `Edit(//Users/<user>/.claude/ecc-homunculus/**)` | Fails with the same sensitive-file error; an allow rule does not override the gate |
| Fork CLV2 to pass `--permission-mode` | Not pursued — requires forking a SHA-pinned external and weakens the observer's safety boundary |

The gate is a legitimate protection (an agent should not silently rewrite its own config dir).
CLV2 instincts are learned *runtime state*, not agent configuration, so the fix moves the state
out of the config dir instead of loosening the gate.

## What changed

**Root cause**

- `home/dot_config/zsh/claude.zsh` — derive an account slug from the config dir basename
  (`.claude` → `default`, `.claude-*` → suffix) and point `CLV2_HOMUNCULUS_DIR` at
  `~/.local/share/ecc-homunculus-<slug>`. The per-account suffix keeps `cld` and `cld-r06`
  isolated from each other and both distinct from the bare-`claude` fallback
  (`~/.local/share/ecc-homunculus`), which is unchanged. `XDG_DATA_HOME` is deliberately not
  consulted so this and `morning-radar.sh` cannot drift apart through two precedence chains.

**The symptoms #336 originally reported**

- Same file — `OBSERVER_ACTIVE_HOURS_START` / `OBSERVER_ACTIVE_HOURS_END` default to `0`, which
  disables the session-guardian clock gate (upstream default 800–2300). That gate skipped ~90
  analysis cycles per project because sessions here routinely run past midnight. The guardian's
  cooldown and idle gates still throttle cycles, so the observer does not get busier while
  nobody is at the keyboard.
- Same file — `ECC_OBSERVER_MAX_TURNS` defaults to `100`, the upstream cap, instead of the
  auto-scaled floor of 20 that cut the Read → dedup-check → Write pass off mid-write. The
  existing 300 s watchdog, not the turn ceiling, bounds a cycle — and the comment added in #256
  already assumed 100 turns, so the two are now consistent.
- All three follow the existing `${VAR:-default}` form, so an explicit override still wins.

**Keeping every reference consistent**

- `home/run_onchange_after_14-enable-clv2-observer.sh.tmpl` — writes `observer.enabled=true` to
  the new per-account paths. The account config dir still selects *which* accounts are in use;
  only the state location moved. Missing this would have left the observer disabled at the new
  path. It also restricts the state dir to `700`, since `observations.jsonl` records tool
  input/output and now sits in the shared `~/.local/share` tree instead of under the config dir;
  the `chmod` runs every time so an existing dir is healed rather than left at the umask default.
  Its slug derivation mirrors `_claude_with_home` branch for branch, both switching on the
  basename, so the two cannot disagree on a directory name neither of them was written for.
- `home/dot_claude/executable_morning-radar.sh` — the headless morning brief uses the `default`
  slug *and* mirrors the three observer knobs, honouring the existing "keep in sync with
  `_claude_with_home`" contract. Both halves matter: the brief lazy-starts a CLV2 observer through
  the `PreToolUse` hook and that observer inherits the wrapper's environment for its whole
  lifetime, so a wrapper carrying only the path would have kept running its own observer on
  exactly the clock gate and turn floor this PR removes.
- `home/dot_agents/skills/knowledge-distill/SKILL.md` — fallback default updated, plus the Phase 0
  diagnostic now greps for both #336 failure signatures (`Reached max turns`, `sensitive file`)
  and reports them as table rows, so a regression is detectable by diagnosis rather than by
  noticing that instincts never appear.
- `home/dot_claude/executable_clv2-session-notify.sh` — comment only; resolution already goes
  through `CLV2_HOMUNCULUS_DIR` first.
- `home/.chezmoiignore` — the account-dir entries stay, now labelled as covering pre-#336 leftovers
  only, and `.local/share/ecc-homunculus-*` is added alongside them. That path is outside
  chezmoi's target tree today, but `observations.jsonl` holds tool input/output whose redaction is
  best-effort, so in a public repo a stray `chezmoi add` under `~/.local/share` must not be able to
  pick it up.
- `home/dot_claude/executable_statusline.sh` — **not changed**: its XDG/`~/.local/share` branch is
  the bare-`claude` fallback, which remains correct. It reads `CLV2_HOMUNCULUS_DIR` first, so the
  🧬N segment follows the move automatically.
- Docs (EN + JA mirrors): `docs/agents/claude-code.*`, `docs/agents/account-isolation.*`,
  `docs/explanation/secrets-and-isolation.*`, `docs/architecture/lifecycle-scripts.*` — new paths,
  the three new env vars documented, and the design reason recorded (why CLV2 state must sit
  outside the config dir while `CLAUDE_CONFIG_DIR` / `ECC_AGENT_DATA_HOME` /
  `GATEGUARD_STATE_DIR` legitimately stay inside it: their writers bypass the Write tool).

**Tests**

- `tests/zsh_aliases.bats` — pins the real per-account values, then pins the property that made
  them necessary: neither account may resolve under `~/.claude*`, the two must differ, and both
  must differ from the unsuffixed fallback. Also pins the three observer env defaults and their
  overridability.
- `tests/files.bats` — the observer-enable test now asserts the new paths, that nothing is created
  under the config dir, and that the state dir ends up at mode `700` (with a BSD/GNU `stat`
  fallback so it holds on both CI platforms). A new test pins the `morning-radar.sh` sync
  contract by deriving the expected value from `_claude_with_home` itself rather than repeating
  the literal, so reverting that one hand-copied line can no longer pass CI.

The `Write(**/.env*)` deny-rule warning that #336 noted as a secondary observation needs no
action: source and deployed `settings.json` already use the `Edit(...)` form.

## Verification

- `make lint` — pass
- `make test` — pass (lint + 173 bats tests)
- `chezmoi diff` — only the intended lines change in the four deployed files
- Reviewed in two rounds. The first (code quality, security, test coverage) returned APPROVE with
  no blocking findings and contributed the `chmod 700`, the mirrored slug derivation and two extra
  assertions; it also injected three regressions (reverting the path, collapsing both accounts onto
  one slug, reverting only the `r06` branch of the enable script) and confirmed each is caught. The
  second round added a cross-model pass and found the half-honoured sync contract described above —
  the only MUST across either round. Overall security risk: low. The posted review comment lists
  every finding with its resolution, including the four rejected with reasons.

## Migration (run locally after merge)

The observer state itself is runtime data, so no migration script is added: this repo manages
config, not state. Run the move **before** `chezmoi apply`, so the destinations do not exist yet:

```bash
# 0. Confirm the destinations are still absent (expected: both "absent")
for d in ~/.local/share/ecc-homunculus-default ~/.local/share/ecc-homunculus-r06; do
  [ -e "$d" ] && echo "PRESENT: $d" || echo "absent:  $d"
done

# 1. Move the accumulated observations/projects state (~450 KB, 171+ observations)
mv ~/.claude/ecc-homunculus     ~/.local/share/ecc-homunculus-default
mv ~/.claude-r06/ecc-homunculus ~/.local/share/ecc-homunculus-r06

# 2. Deploy: run_onchange_after_14 re-runs and merges observer.enabled=true into the moved config
chezmoi apply -v

# 3. Pick up the new alias env, then retire the observers still holding the old one
exec zsh
~/.agents/skills/continuous-learning-v2/agents/start-observer.sh stop || true
```

If step 0 reports `PRESENT` because `chezmoi apply` already ran, those dirs hold only a generated
`config.json`; copy the history in over it instead of moving
(`rsync -a ~/.claude/ecc-homunculus/ ~/.local/share/ecc-homunculus-default/`). The old
`config.json` also carries `observer.enabled: true`, so overwriting the generated one is safe.

## Post-merge verification (issue AC 1)

#336's first acceptance criterion — at least one instinct generated from a real session — needs
the state move plus an observer restart, so it is verified after merge, not in CI:

```bash
# The wrapper points outside the config dir
_claude_with_home "$HOME/.claude" printenv CLV2_HOMUNCULUS_DIR

# The observer is enabled at the new path
jq '.observer' ~/.local/share/ecc-homunculus-default/config.json

# After a cld session has run long enough for an analysis cycle: at least one instinct exists
ls ~/.local/share/ecc-homunculus-default/instincts/personal/ 2>/dev/null
ls ~/.local/share/ecc-homunculus-default/projects/*/instincts/personal/ 2>/dev/null

# The old failure signatures stop appearing
grep -h 'sensitive file\|Reached max turns' \
  ~/.local/share/ecc-homunculus-default/projects/*/observer.log 2>/dev/null | tail -5
```

closes #336

### DIFF
--- a/docs/agents/account-isolation.ja.md
+++ b/docs/agents/account-isolation.ja.md
@@ -18,12 +18,14 @@
 |---|---|---|
 | `CLAUDE_CONFIG_DIR` | `~/.claude` | `~/.claude-r06` |
 | `ECC_AGENT_DATA_HOME` | `~/.claude` | `~/.claude-r06` |
-| `CLV2_HOMUNCULUS_DIR` | `~/.claude/ecc-homunculus` | `~/.claude-r06/ecc-homunculus` |
+| `CLV2_HOMUNCULUS_DIR` | `~/.local/share/ecc-homunculus-default` | `~/.local/share/ecc-homunculus-r06` |
 | `ECC_MCP_HEALTH_STATE_PATH` | `~/.claude/mcp-health-cache.json` | `~/.claude-r06/mcp-health-cache.json` |
 | `GATEGUARD_STATE_DIR` | `~/.claude/.gateguard` | `~/.claude-r06/.gateguard` |
 | `CODEX_HOME` | （デフォルト — `~/.codex`） | `~/.codex-r06` |
 
 r06 の Claude 設定ディレクトリ（`~/.claude-r06`）には、すべての設定アーティファクト（settings、agents、commands、skills）が `~/.claude` を指すシンボリックリンクのみが含まれます。アカウント間で異なるのは、これらの環境変数がツールに書き込むよう指示するランタイム状態のみです。
+
+`CLV2_HOMUNCULUS_DIR` だけは `~/.claude` / `~/.claude-r06` の外、`~/.local/share/` 配下に完全に置かれています。Claude Code は config dir 配下のすべてのパスを sensitive file として扱い Write に対話承認を要求しますが、CLV2 の解析セッションは headless（`claude --model haiku --print`）で動くためその承認を得ることができず、instinct の書き込みがすべて失敗していました（issue #336）。他のアカウント別変数（`CLAUDE_CONFIG_DIR`、`ECC_AGENT_DATA_HOME`、`ECC_MCP_HEALTH_STATE_PATH`、`GATEGUARD_STATE_DIR`）が config dir 配下のままなのは、それらが Claude Code の Write ツールを経由せず node / shell コードから直接書き込まれるため、このゲートに引っかからないからです。
 
 ---
 
@@ -67,9 +69,16 @@ _claude_with_home() {
   local home_dir="$1"
   shift
   (($#)) || set -- claude
+  local account_dir_name="${home_dir:t}"
+  local homunculus_slug
+  case "$account_dir_name" in
+    .claude) homunculus_slug=default ;;
+    .claude-*) homunculus_slug="${account_dir_name#.claude-}" ;;
+    *) homunculus_slug="$account_dir_name" ;;
+  esac
   CLAUDE_CONFIG_DIR="$home_dir" \
     ECC_AGENT_DATA_HOME="$home_dir" \
-    CLV2_HOMUNCULUS_DIR="$home_dir/ecc-homunculus" \
+    CLV2_HOMUNCULUS_DIR="$HOME/.local/share/ecc-homunculus-${homunculus_slug}" \
     ECC_MCP_HEALTH_STATE_PATH="$home_dir/mcp-health-cache.json" \
     GATEGUARD_STATE_DIR="$home_dir/.gateguard" \
     EXA_API_KEY="${EXA_API_KEY:-}" \
--- a/docs/agents/account-isolation.ja.md
+++ b/docs/agents/account-isolation.ja.md
@@ -74,7 +74,7 @@ _claude_with_home() {
   case "$account_dir_name" in
     .claude) homunculus_slug=default ;;
     .claude-*) homunculus_slug="${account_dir_name#.claude-}" ;;
-    *) homunculus_slug="$account_dir_name" ;;
+    *) homunculus_slug="${account_dir_name#.}" ;;
   esac
   CLAUDE_CONFIG_DIR="$home_dir" \
     ECC_AGENT_DATA_HOME="$home_dir" \
--- a/docs/agents/account-isolation.md
+++ b/docs/agents/account-isolation.md
@@ -74,7 +74,7 @@ _claude_with_home() {
   case "$account_dir_name" in
     .claude) homunculus_slug=default ;;
     .claude-*) homunculus_slug="${account_dir_name#.claude-}" ;;
-    *) homunculus_slug="$account_dir_name" ;;
+    *) homunculus_slug="${account_dir_name#.}" ;;
   esac
   CLAUDE_CONFIG_DIR="$home_dir" \
     ECC_AGENT_DATA_HOME="$home_dir" \
--- a/docs/agents/account-isolation.md
+++ b/docs/agents/account-isolation.md
@@ -18,12 +18,14 @@ These variables are set inline on the agent subprocess — they are never export
 |---|---|---|
 | `CLAUDE_CONFIG_DIR` | `~/.claude` | `~/.claude-r06` |
 | `ECC_AGENT_DATA_HOME` | `~/.claude` | `~/.claude-r06` |
-| `CLV2_HOMUNCULUS_DIR` | `~/.claude/ecc-homunculus` | `~/.claude-r06/ecc-homunculus` |
+| `CLV2_HOMUNCULUS_DIR` | `~/.local/share/ecc-homunculus-default` | `~/.local/share/ecc-homunculus-r06` |
 | `ECC_MCP_HEALTH_STATE_PATH` | `~/.claude/mcp-health-cache.json` | `~/.claude-r06/mcp-health-cache.json` |
 | `GATEGUARD_STATE_DIR` | `~/.claude/.gateguard` | `~/.claude-r06/.gateguard` |
 | `CODEX_HOME` | (default — `~/.codex`) | `~/.codex-r06` |
 
 The r06 Claude config directory (`~/.claude-r06`) contains only symlinks pointing back to `~/.claude` for every config artifact (settings, agents, commands, skills). What differs between accounts is entirely in the state that these env vars direct the tools to write.
+
+`CLV2_HOMUNCULUS_DIR` is the one directory that lives entirely outside `~/.claude` / `~/.claude-r06`, under `~/.local/share/`: Claude Code treats every path under the config dir as a sensitive file requiring interactive Write approval, and the CLV2 analysis session runs headless (`claude --model haiku --print`), so it could never grant that approval and every instinct write failed (issue #336). The other per-account vars (`CLAUDE_CONFIG_DIR`, `ECC_AGENT_DATA_HOME`, `ECC_MCP_HEALTH_STATE_PATH`, `GATEGUARD_STATE_DIR`) stay under the config dir because they are written directly by node/shell code rather than through Claude Code's Write tool, so they never hit that gate.
 
 ---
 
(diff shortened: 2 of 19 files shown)


kryota-dev (commented):
## Multi-review summary

Three independent reviewers ran against `3441d68`: a general code-quality pass, a security pass,
and a cross-model pass (Codex). Every finding below was either fixed in `2a32d44` or rejected with
a reason — nothing was left open, so no inline comments are attached.

| Category | code-quality | security | codex |
|---|---|---|---|
| MUST | 0 | 0 | 1 |
| SHOULD | 1 | 1 | 2 |
| NITS | 3 | 4 | 3 |
| GOOD | 3 | 6 | 3 |

Security: overall risk **low**. No change to `settings.json` `permissions`, the gateguard, or
`ECC_DISABLED_HOOKS` — the sensitive-file gate is stepped around by moving state out of its scope,
never loosened.

### Fixed in `2a32d44`

**[MUST] The "keep in sync" contract was only half honoured** (`executable_morning-radar.sh`,
codex, confidence 0.99). The wrapper carried the relocated storage path but not
`OBSERVER_ACTIVE_HOURS_START`/`END` or `ECC_OBSERVER_MAX_TURNS`. Verified: the brief lazy-starts a
CLV2 observer through the `PreToolUse` hook, and that observer inherits the wrapper's environment
for its whole process lifetime — so the launchd brief would have kept running its observer on
exactly the upstream clock gate and auto-scaled turn floor this PR exists to get past, while
interactive sessions were fixed. All three are now mirrored, and the sync test derives all four
values from the zsh helper instead of restating them.

**[SHOULD] Diagnostic and `evolve` could read different stores**
(`knowledge-distill/SKILL.md`, codex, 0.98). With `CLV2_HOMUNCULUS_DIR` unset, Phase 0 diagnosed
`ecc-homunculus-default` while `instinct-cli.py evolve` resolved the upstream bare-`claude`
fallback. Pre-existing rather than introduced here, but one token to fix: the CLI call now gets
`CLV2_HOMUNCULUS_DIR="$H"`.

**[SHOULD] `.chezmoiignore` did not cover the new location** (security, medium). The old path was
listed; the new one was not, on the reasoning that `~/.local/share` is outside chezmoi's target
tree. That is true today but not a guarantee, and `observations.jsonl` holds tool input/output
whose redaction is best-effort — in a public repo the entry is cheap insurance, so
`.local/share/ecc-homunculus-*` is now ignored too.

**[SHOULD] Docs snippet drifted from the implementation** (`account-isolation.md` / `.ja.md`,
raised independently by all three reviewers). The wildcard branch still showed
`homunculus_slug="$account_dir_name"` while the code had moved to `${account_dir_name#.}` — the
snippet was copied before that change. Corrected in both mirrors.

**[NITS] A docs claim went stale** (`ci-and-tests.md` / `.ja.md`, codex, 1.00). Both mirrors stated
that no bats test asserts `CLV2_HOMUNCULUS_DIR`; one now does. Rewritten. This one is worth
calling out because it does not contain the string `ecc-homunculus`, so the path-based sweep that
found every other site could not have caught it.

**[NITS] `.claude-default` would collide** (security, low; same theme as codex's "only two
profiles" and code-quality's empty-suffix note). Unreachable today, but a `.claude-default` account
dir would derive the slug `default` and silently merge two accounts' observations. Documented as a
warning next to the mapping rather than guarded in code, since no caller can reach it.

**[NITS] The privacy note overclaimed** (security, info). `run_onchange_after_14` said "nothing
leaves the machine", which is true of the recording step but not the analysis step — the observer
feeds batches to `claude --model haiku --print`. Now stated plainly, which also matters because
disabling the clock gate widens when that happens.

### Rejected

**[SHOULD] Make the morning-radar test behavioural rather than source-text** (codex, 0.97).
Partially taken: the test still greps, but now derives all four expected values from
`_claude_with_home`, so it fails if either side drifts. Full behavioural execution was not added —
the wrapper would have to be run with a fake `claude` and a day-guard stub, and the residual gap
it would close (an assignment kept but overridden later in the file) is already covered indirectly
by the behavioural test on the helper itself.

**[NITS] `chmod` follows symlinks and there is a window between `mkdir -p` and `chmod`**
(security, info, 0.35 confidence). Not a regression — the same pattern predates this PR — and the
threat model is a single-user macOS workstation, where an attacker able to pre-create that symlink
already runs as the same user.

**[NITS] No cleanup of the old `<account>/ecc-homunculus` directories** (code-quality, medium).
The migration in the PR body moves those directories rather than leaving them, so following it
leaves nothing orphaned. Deliberately not automated: this repo manages config, not runtime state.

**[NITS] "Claude Code classifies every path under the config dir as a sensitive file" is an
unverified generalisation** (code-quality, low confidence on the objection itself). Kept as
written. The measurement behind it used two arbitrary new filenames at two different depths — the
config-dir root and a nested `instincts/personal/` path — both refused with `... which is a
sensitive file`, while arbitrary paths outside the dir were written without a prompt. That is
prefix-scoped behaviour, not a per-file classification.

### Called out as done well

- Not weakening the gate: relocating the state instead of adding an allow rule or forking the
  pinned external keeps the protection that stops an agent rewriting its own config dir, and the
  allow-rule route was measured to fail anyway.
- `chmod 700` on the state dir, given the move into a tree shared with other tools.
- The regression guards encode the defect class, not the happy path: the storage dir may not
  resolve under the config dir, and the two accounts plus the bare-`claude` fallback must stay
  mutually distinct. The test review injected three regressions (reverting the path, collapsing
  both accounts onto one slug, reverting only the `r06` branch of the enable script) and confirmed
  each is caught.
- Leaving `statusline.sh` untouched was independently confirmed correct by two reviewers: it
  consults `CLV2_HOMUNCULUS_DIR` first, and its XDG fallback belongs to bare `claude`.
- Tests sandbox `HOME` and `XDG_DATA_HOME`, so they never touch real observer state.

PR #350, opened 2026-07-26
feat: replace desktop-notify with a self-hosted ntfy notification system over Tailscale

kryota-dev:
## Summary

Replaces the ECC `stop:desktop-notify` Stop hook (local osascript, no history) with a
self-hosted **ntfy** notification system reachable over **Tailscale** (closes #337).
Claude Code events now push to tailnet devices with full attribution
(repo / branch / account / session / event) and a persistent, filterable server-side
history. The finalized decision record is committed as
`.claude/prds/337-ntfy-tailscale.prd.md`.

## Key design decisions

- **Docker, not Homebrew**: the Homebrew `ntfy` formula builds macOS binaries with the
  `noserver` Go tag — `ntfy serve` does not exist in the installed binary (verified on
  the target machine). The server therefore runs as the official `binwiederhier/ntfy`
  image (exact tag + digest pin, Renovate-tracked via a compose regex manager) under
  the Brewfile-provisioned Docker Desktop with `restart: unless-stopped`.
- **Network boundary**: the container publishes `127.0.0.1:<port>:80` only; tailnet
  exposure goes exclusively through `tailscale serve --bg` (Tailscale-managed TLS,
  persists across reboots). `tailscale funnel` is prohibited and bats-asserted absent.
  `upstream-base-url: https://ntfy.sh` enables iOS instant push (user-approved
  metadata-only egress; behavior on a tailnet-only server is recorded by the smoke
  test).
- **Auth**: `auth-default-access: deny-all`; write-only publisher token vs read-only
  subscriber token, both stored in 1Password (`Dotfiles - ntfy`), publisher token
  rendered into a 0600 env file via `onepasswordRead`.
- **Fail-open hook wrapper** (`~/.claude/ntfy-notify.sh`): every exit path is 0;
  unreachable server ⇒ log + content-free local alert; no env file ⇒ silent no-op.
  Stop bodies are truncated to 200 codepoints (jq slice, multibyte-safe) and scrubbed
  with the deployed client-identifier pattern. The token only ever travels via a
  `curl -K` config file — never argv/stdout/stderr/logs (bats-asserted).
- **Runtime state stays out of chezmoi**: `user.db` / `cache.db` live in
  `~/Library/Application Support/ntfy` (0700); recovery runbook documented.
- **Intentional deviation from lifecycle convention #6**: `run_onchange_after_31`
  warns and exits 0 when Docker Desktop is down — notifications must never block
  `chezmoi apply` (documented in-script and in docs).

## Changes

- `home/dot_config/ntfy/`: compose (pinned image, loopback ports), `server.yml`
  (cache 168h from the `[ntfy]` SSOT block, deny-all auth, upstream relay), publisher
  env — the two `onepasswordRead` templates are added to the CI exclude list in both
  jobs
- `home/dot_claude/`: `ntfy-notify.sh` wrapper; `settings.json` Notification
  (`permission_prompt|idle_prompt|agent_needs_input|agent_completed`) + Stop hooks,
  `stop:desktop-notify` disabled via `env.ECC_DISABLED_HOOKS` (one-step rollback
  documented)
- `home/run_onchange_after_31-setup-ntfy.sh.tmpl`: state dir, `docker compose up -d`,
  `tailscale serve --bg` assertion (embedded-hash, CI guard)
- Tests: `tests/ntfy_notify.bats` (14 cases), files.bats ntfy section (9 asserts),
  docs_facts constants; **`make test`: 206 ok / 0 failures**
- Docs: new `docs/architecture/notifications.md` (+ja) with setup / smoke-test /
  recovery / rollback runbooks; lifecycle, secrets, CI docs and FACT markers synced
  (item count 4→6, exclusion count 6→8)

## Review

Deliberated via a 4-perspective council (UX / engineering / security / ops) with a
2-reviewer adversarial pass on the PRD, then implemented and re-reviewed by a
3-role review team (code-quality / security / test) — all APPROVE. Notable
review-driven fixes: multibyte truncation moved from locale-dependent `cut -c` to jq
codepoint slicing; topic-name fallbacks removed in favor of the SSOT; attribution
and fail-open branches now bats-covered.

## Post-merge setup (one-time, documented in the runbook)

1. Create the `Dotfiles - ntfy` 1Password item (`base-url`, `credential`).
2. `chezmoi apply`, then provision users/tokens via `docker compose exec` and store
   them in 1Password.
3. Subscribe devices with the read-only token; run the documented smoke test
   (anonymous denial, ro-publish denial, iOS instant-push observation).

### DIFF
--- a/docs/architecture/notifications.md
+++ b/docs/architecture/notifications.md
@@ -1,0 +1,130 @@
+# Notifications: self-hosted ntfy over Tailscale
+
+[日本語版](notifications.ja.md)
+
+Claude Code session events are pushed to a ntfy server self-hosted on this Mac and
+subscribed from tailnet devices (kryota-dev/dotfiles#337). This replaces the ECC
+`stop:desktop-notify` Stop hook (local osascript, no history) with attributable,
+persistent, remotely subscribable notifications. The finalized decision record lives
+in `.claude/prds/337-ntfy-tailscale.prd.md`.
+
+## Architecture
+
+```
+Claude Code hooks (settings.json)
+  Notification: permission_prompt | idle_prompt | agent_needs_input | agent_completed
+  Stop
+        │ stdin JSON (async, fail-open)
+        ▼
+~/.claude/ntfy-notify.sh ──── enrich (repo/branch/account/session/event)
+        │ curl -K, Bearer write-only token, --max-time 3
+        ▼
+http://127.0.0.1:2586 ── docker: binary/ntfy (restart: unless-stopped)
+        ▲                        └─ state: ~/Library/Application Support/ntfy/
+tailscale serve --bg (HTTPS, MagicDNS)          (user.db, cache.db — outside chezmoi)
+        ▲
+phones/tablets on the tailnet (read-only token)
+```
+
+- **Server runtime**: the Homebrew `ntfy` formula builds macOS binaries with the
+  `noserver` tag (`ntfy serve` does not exist), so the server runs as the official
+  `binary/ntfy` Docker image under Docker Desktop. The image tag is pinned in
+  `home/dot_config/ntfy/compose.yaml.tmpl` and tracked by Renovate.
+- **Network boundary**: the container binds `127.0.0.1` only; tailnet exposure goes
+  exclusively through `tailscale serve --bg` (persists across reboots). `tailscale
+  funnel` is prohibited — verify with `tailscale funnel status` (expect nothing
+  served) whenever in doubt.
+- **iOS instant push**: `upstream-base-url: https://ntfy.sh` relays poll requests
+  (topic metadata only, never message bodies) through ntfy.sh/APNs — the sole
+  approved metadata egress. The device still fetches bodies from this server over
+  the tailnet. Record the observed iOS behavior here after the smoke test:
+  _observed: (pending first smoke test)_.
+- **Topics** (fixed, urgency-based; SSOT in `home/.chezmoidata.toml` `[ntfy]`):
+
+| Topic | Events | Priority | Intended device setting |
+|-------|--------|----------|------------------------|
+| `claude-attention` | permission_prompt, idle_prompt, agent_needs_input | high (4) | sound/vibrate on |
+| `claude-done` | agent_completed, Stop | default (3) | silent delivery |
+| `claude-test` | manual smoke tests | — | mute after testing |
+
+Attribution (repo, branch, account `default`/`r06`, 8-char session id, event type)
+travels in titles and tags, never in topic names. Stop bodies are truncated to 200
+chars and scrubbed with the client-identifier pattern from
+`~/.config/git/gitleaks-own.toml` (best-effort name scrub — **not** a secret/PII
+detector; truncation is the primary defense). Accepted residual risk (#337 PRD): a
+prompt-injected assistant message could still surface short secret fragments within
+the 200-char window — general secret detection is explicitly out of scope.
+
+## Setup runbook (one-time)
+
+1. **1Password item** — create `Dotfiles - ntfy` in the `kryota.dev` vault with
+   fields `base-url` (the serve endpoint, `https://<host>.<tailnet>.ts.net`) and
+   `credential` (placeholder until step 4). See
+   [secrets-1password](../getting-started/secrets-1password.md).
+2. **Docker Desktop** — enable *Start Docker Desktop when you sign in* (Settings →
+   General); the compose `restart: unless-stopped` policy then revives the server
+   after every reboot.
+3. **Apply** — `chezmoi apply` renders the config, starts the container
+   (`run_onchange_after_31-setup-ntfy`), and asserts the `tailscale serve --bg`
+   mapping.
+4. **Provision auth** (inside the container; auth is runtime state, never chezmoi-managed):
+
+   ```bash
+   cd ~/.config/ntfy
+   docker compose exec ntfy ntfy user add publisher     # publisher, write-only
+   docker compose exec ntfy ntfy user add subscriber    # devices, read-only
+   for t in claude-attention claude-done claude-test; do
+     docker compose exec ntfy ntfy access publisher "$t" write-only
+     docker compose exec ntfy ntfy access subscriber "$t" read-only
+   done
+   docker compose exec ntfy ntfy token add --label chezmoi publisher
+   docker compose exec ntfy ntfy token add --label devices subscriber
+   ```
+
+   Store the publisher token in the item's `credential` field, the subscriber token
+   in a separate field (entered on devices manually), then re-run `chezmoi apply`.
+5. **Subscribe devices** — ntfy app (iOS/Android) → *Use another server* with the
+   `base-url` endpoint, the subscriber token from 1Password, and the two topics.
+
+## Smoke test
+
+```bash
+# anonymous publish must be denied (401/403)
+curl -s -o /dev/null -w '%{http_code}\n' -d test "$(op read 'op://kryota.dev/Dotfiles - ntfy/base-url')/claude-test"
+# read-only token must be denied for publish (403)
+curl -s -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer <subscriber-token>" -d test ".../claude-test"
+# publisher token succeeds (200); phones should receive it
+curl -s -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer <publisher-token>" -d test ".../claude-test"
+# history retrieval + filtering examples (verify once, then rely on them)
+curl -s -H "Authorization: Bearer <subscriber-token>" ".../claude-done/json?poll=1&since=24h" | jq 'select(.tags | index("dotfiles"))'
+curl -s -H "Authorization: Bearer <subscriber-token>" ".../claude-attention/json?poll=1&since=all" | jq 'select(.tags | index("permission_prompt"))'
+```
+
+Also record whether iOS delivers instantly with the app closed (upstream relay on a
+tailnet-only server is unverified upstream — see the PRD).
+
+## Failure modes & troubleshooting
+
+| Symptom | Cause / fix |
+|---------|-------------|
+| Local alert sound, no phone notification | Wrapper publish failed — server down. Check `~/Library/Logs/ntfy-notify.log`, then `docker compose -f ~/.config/ntfy/compose.yaml up -d` |
+| No notifications right after login | Docker Desktop still starting; the fail-open window is expected. Enable start-at-login (runbook step 2) |
+| `chezmoi apply` prints `[ntfy] Docker Desktop is not running` | Intentional warn-and-skip (deviation from lifecycle convention #6): notifications must not block apply. Start Docker Desktop, re-run apply or the printed compose command |
+| Phone can't reach the server | Device off the tailnet, or `tailscale serve` mapping lost — re-run `tailscale serve --bg http://127.0.0.1:2586` (also asserted by every re-triggered apply) |
+| Old messages missing | `cache-duration` (168h, `[ntfy]` in `.chezmoidata.toml`) elapsed |
+| Notifications on the wrong account badge | `CLAUDE_CONFIG_DIR` unset in that session; account falls back to `default` |
+
+## Recovery: auth.db lost or corrupted
+
+Delete `~/Library/Application Support/ntfy/user.db` if corrupted, then re-run
+runbook step 4. **Re-issuing tokens invalidates every existing token**: update the
+1Password `credential` field, re-run `chezmoi apply`, and re-enter the subscriber
+token on every subscribed device.
+
+## Rollback (one step)
+
+Revert the two settings.json changes together — remove `stop:desktop-notify` from
+`env.ECC_DISABLED_HOOKS` **and** delete the `notification:ntfy-notify` /
+`stop:ntfy-notify` hook entries — then `chezmoi apply`. Doing only one half either
+leaves you notification-less or double-notifies. Server teardown (optional):
+`docker compose -f ~/.config/ntfy/compose.yaml down` and `tailscale serve --https=443 off`.
--- a/docs/architecture/notifications.md
+++ b/docs/architecture/notifications.md
@@ -8,6 +8,13 @@ subscribed from tailnet devices (kryota-dev/dotfiles#337). This replaces the ECC
 persistent, remotely subscribable notifications. The finalized decision record lives
 in `.claude/prds/337-ntfy-tailscale.prd.md`.
 
+**Scope boundary**: this page covers only the Claude Code `Notification`/`Stop`
+hook → ntfy path. The repo has three other, independent local-only notification
+paths that this system deliberately does not touch: `clv2-session-notify.sh`
+(SessionStart, instinct-cluster review nudge), `morning-radar.sh` (launchd,
+osascript morning brief), and the `notify` zsh alias (audible chime, also reused
+by this system's wrapper as its failure alert sound).
+
 ## Architecture
 
 ```
@@ -19,7 +26,7 @@ Claude Code hooks (settings.json)
 ~/.claude/ntfy-notify.sh ──── enrich (repo/branch/account/session/event)
         │ curl -K, Bearer write-only token, --max-time 3
         ▼
-http://127.0.0.1:2586 ── docker: binary/ntfy (restart: unless-stopped)
+http://127.0.0.1:2586 ── docker: binwiederhier/ntfy (restart: unless-stopped)
         ▲                        └─ state: ~/Library/Application Support/ntfy/
 tailscale serve --bg (HTTPS, MagicDNS)          (user.db, cache.db — outside chezmoi)
         ▲
@@ -28,16 +35,19 @@ phones/tablets on the tailnet (read-only token)
 
 - **Server runtime**: the Homebrew `ntfy` formula builds macOS binaries with the
   `noserver` tag (`ntfy serve` does not exist), so the server runs as the official
-  `binary/ntfy` Docker image under Docker Desktop. The image tag is pinned in
+  `binwiederhier/ntfy` Docker image under Docker Desktop. The image tag is pinned in
   `home/dot_config/ntfy/compose.yaml.tmpl` and tracked by Renovate.
 - **Network boundary**: the container binds `127.0.0.1` only; tailnet exposure goes
   exclusively through `tailscale serve --bg` (persists across reboots). `tailscale
   funnel` is prohibited — verify with `tailscale funnel status` (expect nothing
   served) whenever in doubt.
-- **iOS instant push**: `upstream-base-url: https://ntfy.sh` relays poll requests
-  (topic metadata only, never message bodies) through ntfy.sh/APNs — the sole
-  approved metadata egress. The device still fetches bodies from this server over
-  the tailnet. Record the observed iOS behavior here after the smoke test:
+- **iOS instant push**: `upstream-base-url: https://ntfy.sh` publishes a poll
+  request to ntfy.sh for **every incoming message** — regardless of whether any
+  iOS device subscribes (topic metadata only, never message bodies) — feeding
+  APNs for instant delivery. This is the sole approved metadata egress. The
+  device still fetches bodies from this server over the tailnet. If the smoke
+  test shows instant push does not work tailnet-only, remove the key to bring
+  egress to zero. Record the observed iOS behavior here after the smoke test:
   _observed: (pending first smoke test)_.
 - **Topics** (fixed, urgency-based; SSOT in `home/.chezmoidata.toml` `[ntfy]`):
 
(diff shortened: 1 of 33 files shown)


kryota-dev (commented):
## Multi-Review integrated result (cc-code-review / cc-security-review / codex / architecture-reviewer)

### Verdict summary

| Category | cc-code-review | cc-security-review | codex | architecture-reviewer |
|---|---|---|---|---|
| MUST / High | 1 | 1 (High) | 2 (P1) | 1 |
| SHOULD / Med+Low | 4 | 10 (2 Med, 8 Low) | 0 | 6 |
| NITS / Info | 2 | 6 (Info) | 0 | 1 |
| GOOD | 5 | 11 | 0 | 2 |

Security overall risk: **High** until the image reference is fixed; the remaining design (loopback bind + `tailscale serve`, deny-all + write-only/read-only token split, `curl -K` token hygiene, runtime state outside the chezmoi tree) was assessed as solid by all reviewers.

### [MUST] (2 after adversarial verification)

1. **`binary/ntfy` does not exist — the official image is `binwiederhier/ntfy`** (found independently by 3 reviewers; **verified by direct Docker Hub API observation**: `binary/ntfy` → `object not found`, `binwiederhier/ntfy:v2.26.0` → exists, multi-arch). The wrong name propagates consistently to `compose.yaml.tmpl`, `renovate.json5`, two `files.bats` assertions, `notifications.md`/`.ja.md`, and the PRD — and the bats suite asserts the wrong literal, so `make test` cannot catch it. Beyond availability (first `docker compose up` fails), this is a supply-chain hazard: the `binary` Docker Hub account exists (0 public repos) and could later publish a malicious `ntfy` tag that this setup would pull unverified. Fix all sites together and consider a digest pin. → inline comment on `compose.yaml.tmpl:12`.
2. **`docs/agents/claude-code.md`/`.ja.md` (the canonical hooks-graph doc) is not updated** (architecture-reviewer; verified by grep: zero `ntfy` mentions; `stop:desktop-notify` still listed as active at the mermaid graph and the Stop table, and the new top-level `Notification` hook type is absent). The doc that mirrors `settings.json` as primary material now drifts from it. Not inline-commentable (file not in the diff) — fix in this PR.

### Adversarial verification log (large-tier protocol)

- M1 (image name): confirmed by the parent via direct primary-source observation (Docker Hub API); no refutation round needed for a directly observed fact.
- M2 (codex P1: warn-exit-0 poisons `run_onchange` retry state, breaking the documented "re-run apply" recovery): 3 independent adversarial verifiers with distinct lenses. **Correctness lens: CONFIRMED (0.95)** — chezmoi v2.71.1 source (`TargetStateScript.Apply`) records state only on exit 0; **Reproduction lens: CONFIRMED (0.95)** — empirically reproduced with a sandbox chezmoi source (2nd apply is a no-op after an exit-0 first run); **Impact lens: REFUTED as MUST (0.75)** — impact is bounded to the first-provisioning window (`restart: unless-stopped` + start-at-login auto-recovers afterwards), the printed manual `docker compose up -d` recovery does work, and warn-exit-0 matches the established repo convention used by scripts 18/40/50. **Adjudication: downgraded to SHOULD** — fix the inaccurate "re-run `chezmoi apply`" wording (script comment + notifications.md/.ja + lifecycle-scripts.md/.ja) and make `docker compose up` failure itself warn-and-skip for fail-open consistency. → inline comment on `run_onchange_after_31-setup-ntfy.sh.tmpl:39`.
- M3 (hooks-doc drift): confirmed by grep (directly observed absence); no refutation round needed.

### [SHOULD] highlights (inline comments posted for each)

- scrub/truncate order lets identifiers straddle the 200-char boundary and can exceed the cap after `[redacted]` substitution (reproduced: 202 chars) — swap to scrub-then-truncate with a pre-cap.
- `title`/`tags` (repo, branch) bypass the client-identifier scrub — the fields most likely to carry client names; on Notification events they are the only content.
- `trap ... EXIT` misses signal kills; the hook's 10s timeout can strand a 0600 curl config containing the Bearer token in TMPDIR — extend to `EXIT INT TERM HUP`.
- `NTFY_NOTIFY_ENV_FILE` override makes `. "$ENV_FILE"` an arbitrary-code sink if a hostile project env reaches hook processes (unverified propagation; low likelihood) — validate ownership/permissions before sourcing, or drop the override.
- `upstream-base-url` sends a poll request for **every** message, not just when iOS subscribes (docs.ntfy.sh/config verified) — current "iOS instant push" framing understates the egress; fix the doc wording, consider `upstream-access-token` as a follow-up.
- New 1Password references lack the content validation the Redact Patterns item already gets (YAML/squote injection hardening).
- `topic_test` is declared in the `[ntfy]` SSOT but referenced by nothing (docs hardcode `claude-test`).
- Runbook/smoke-test examples pass tokens via `-H`/argv — the exact pattern the wrapper's own bats tests forbid.
- Architecture: link `notifications.md` from `docs/architecture/overview.md`'s subsystem table; state its scope boundary vs the other notification paths (clv2-session-notify, morning-radar, notify alias); note the Docker Desktop runtime dependency in `dev-tooling.md`.
- Current hooks docs claim: `Notification` events now document a `message` field, and `agent_needs_input`/`agent_completed` are documented to fire only while an agent view is open — both affect D3's coverage assumptions; verify during the smoke test and record the outcome (medium confidence, flagged 要検証 by the reviewer).

### [NITS]

`auth.db` heading vs actual `user.db` filename in docs; no `command -v docker` before `docker info` (message accuracy); `.chezmoiignore` comment block mixes VS Code and ntfy rationale; no log rotation for `ntfy-notify.log`; consider `name:` in compose to scope `--remove-orphans`; perl scrub has no internal timeout (ReDoS-hardening: `alarm 2`).

### [GOOD] (consensus highlights)

Token hygiene machine-enforced (curl `-K`, argv/log capture tests, hardened static greps); fully injection-safe JSON construction via `jq --arg`; codepoint-safe UTF-8 truncation with a corruption test; runtime state cleanly outside the chezmoi tree (consistent with the CLV2 homunculus precedent); CI exclude-list + FACT-marker discipline followed exactly (both jobs, cross-checked URIs); fail-open contract with a content-free local alert; tailnet MagicDNS name isolated in 1Password.

### Existing-review dedup check

No prior reviews or comments existed on this PR at review time — nothing to dedup. (Note: the `Setup Validation` workflow is `disabled_manually` repo-wide, so AC-011's CI-level verification did not actually run for this PR; local `make test` = 203 ok.)

### Rejected / downgraded findings ledger

| Finding | Verdict | Grounds |
|---|---|---|
| codex P1 #2 "documented recovery never runs" as MUST | Downgraded to SHOULD | 2/3 lenses confirm the mechanism (chezmoi source + empirical repro), but impact lens bounds severity: first-provisioning window only, working printed manual recovery, established repo convention (scripts 18/40/50). Wording fix + compose-failure warn-skip retained as SHOULD. |
| cc-security Medium "env-override code execution" as Medium-blocking | Kept, flagged unverified | Sink is real; the propagation path (project `settings.json` env → hook process) is unverified. Posted as SHOULD with verification note rather than silently dropped (coverage-first). |

home/dot_config/ntfy/compose.yaml.tmpl
```diff
@@ -0,0 +1,24 @@
...
+# formula builds macOS binaries with the `noserver` tag — `ntfy serve` does not
+# exist outside containers/Linux. Managed by run_onchange_after_31-setup-ntfy.
+#
+# The image line below is intentionally literal (no template variables): the
+# Renovate regex manager in .github/renovate.json5 tracks the pinned tag, and
+# tests/files.bats asserts the pin format. Everything else templated here comes
+# from the [ntfy] SSOT block in home/.chezmoidata.toml.
+services:
+  ntfy:
+    image: binary/ntfy:v2.26.0
```
kryota-dev: [MUST] `binary/ntfy` does not exist on Docker Hub — the official ntfy image is `binwiederhier/ntfy` (verified: Docker Hub API returns `object not found` for `binary/ntfy`, while `binwiederhier/ntfy:v2.26.0` exists; docs.ntfy.sh/install uses `binwiederhier/ntfy`). As written, the first `docker compose up -d` fails, and worse, `binary` is a real unrelated Docker Hub account (0 public repos) whose namespace this file would trust if it ever published an `ntfy` tag — a supply-chain hazard with this container receiving the state bind mount and tailnet exposure.

Fix together (the wrong literal is asserted by the test suite itself, so `make test` cannot catch this): this line, `.github/renovate.json5` (`matchStrings`/`depNameTemplate`), the two `tests/files.bats` assertions, `docs/architecture/notifications.md`/`.ja.md`, and the PRD. Consider adding a digest pin (`@sha256:…`) while at it.

home/run_onchange_after_31-setup-ntfy.sh.tmpl
```diff
@@ -0,0 +1,61 @@
...
+# notifications are not setup-critical. Recovery: start Docker Desktop and
+# re-run `chezmoi apply`, or run the compose command below manually.
+if ! docker info >/dev/null 2>&1; then
+  echo "[ntfy] Docker Desktop is not running; skipping server start." >&2
+  echo "[ntfy] Recover with: docker compose -f ~/.config/ntfy/compose.yaml up -d" >&2
+  exit 0
+fi
+
+echo "[ntfy] starting ntfy server (docker compose up -d)..."
+docker compose -f "$HOME/.config/ntfy/compose.yaml" up -d --remove-orphans
```
kryota-dev: [SHOULD] Two related issues with the fail-open contract (severity adjudicated via a 3-lens adversarial verification round — chezmoi-source and empirical lenses confirmed the mechanism, impact lens bounded the severity):

1. `docker compose up -d` failure (image pull error, port conflict, compose syntax) is NOT covered by the warn-and-skip — `set -euo pipefail` makes it hard-fail `chezmoi apply`, contradicting the documented "notifications must never block apply" contract. Wrap it: `if ! docker compose …; then echo warn; exit 0; fi`.
2. The "Recovery: start Docker Desktop and re-run `chezmoi apply`" wording (script comment above + notifications.md troubleshooting + lifecycle-scripts.md) is inaccurate: an exit-0 run records the `run_onchange` state hash, so re-running apply is a no-op until the compose/server templates change (confirmed against chezmoi v2.71.1 source and reproduced empirically). The printed manual `docker compose up -d` command is the recovery path that actually works — make the wording say so.

home/dot_claude/executable_ntfy-notify.sh
```diff
@@ -0,0 +1,193 @@
...
+  [ -n "$cwd" ] || cwd="$PWD"
+  account="$(account_slug)"
+  repo="$(repo_name "$cwd")"
+  branch="$(branch_name "$cwd")"
+
+  title="[${account}] ${repo}@${branch} — ${event}"
+  body=""
+  if [ "$event" = "Stop" ]; then
+    body="$(printf '%s' "$input" | jq -r '.last_assistant_message // empty' 2>/dev/null || true)"
+    body="$(scrub_body "$(truncate_body "$body")")"
```
kryota-dev: [SHOULD] Order of operations: `scrub_body "$(truncate_body …)"` truncates first, so (a) a client identifier straddling the 200-char boundary loses its tail and no longer matches the scrub pattern — its head survives unredacted; (b) `[redacted]` (10 chars) replacing a shorter name pushes the final message over the 200-char cap (reproduced: 202 chars), contradicting AC-007's "truncation always applies". Swap to scrub-then-truncate, with a cheap pre-cap to bound the perl input, e.g. `body="$(truncate_body "$(scrub_body "$(printf '%s' "$body" | head -c 8192)")")"`.

home/dot_claude/executable_ntfy-notify.sh
```diff
@@ -0,0 +1,193 @@
...
+  session_id="$(printf '%s' "$input" | jq -r '.session_id // empty' 2>/dev/null || true)"
+  sid="$(printf '%.8s' "$session_id")"
+  [ -n "$sid" ] || sid='-'
+  cwd="$(printf '%s' "$input" | jq -r '.cwd // empty' 2>/dev/null || true)"
+  [ -n "$cwd" ] || cwd="$PWD"
+  account="$(account_slug)"
+  repo="$(repo_name "$cwd")"
+  branch="$(branch_name "$cwd")"
+
+  title="[${account}] ${repo}@${branch} — ${event}"
```
kryota-dev, line 173: [SHOULD] `repo` and `branch` land in the title and tags unscrubbed. The client-identifier pattern this repo maintains exists precisely because client names show up in repo/branch names (e.g. `acmecorp-web`, `feature/acme-migration`) — and on Notification events, where the body is empty, these are effectively the only content published and cached for 168h. Apply the scrub to the attribution fields too (one combined scrub call keeps hook latency flat).

home/dot_claude/executable_ntfy-notify.sh
```diff
@@ -0,0 +1,193 @@
...
+    --arg sid "$sid" \
+    --argjson priority "$priority" \
+    '{topic: $topic, title: $title, message: $message, priority: $priority,
+      tags: [$emoji, $event, $repo, $account, $sid]}')"
+
+  # Token travels via a 0600 curl config file, never argv.
+  umask 077
+  curl_cfg="$(mktemp "${TMPDIR:-/tmp}/ntfy-notify.XXXXXX")"
+  # shellcheck disable=SC2064
+  trap "rm -f '$curl_cfg'" EXIT
```
kryota-dev: [SHOULD] `trap … EXIT` does not fire when bash dies from an unhandled SIGTERM/SIGINT/SIGHUP. This hook runs with `"timeout": 10`; if Claude Code kills it at the timeout, the 0600 curl config holding the Bearer token stays in `${TMPDIR:-/tmp}` indefinitely. Extend to `trap 'rm -f "$curl_cfg"' EXIT INT TERM HUP` (or switch to process substitution `-K <(printf …)` and never touch the filesystem). A startup sweep of stale `ntfy-notify.*` temp files would add belt-and-suspenders.

home/dot_claude/executable_ntfy-notify.sh
```diff
@@ -0,0 +1,193 @@
...
+main() {
+  local input event_name event topic priority emoji
+  local session_id sid cwd account repo branch title body payload
+  local curl_cfg
+
+  # No env file / no jq -> silent no-op (unprovisioned machine or Linux).
+  [ -f "$ENV_FILE" ] || exit 0
+  command -v jq >/dev/null 2>&1 || exit 0
+  # shellcheck source=/dev/null
+  . "$ENV_FILE"
```
kryota-dev, line 122: [SHOULD] `. "$ENV_FILE"` with an env-var-overridable path is an arbitrary-code-execution sink: anything that can set `NTFY_NOTIFY_ENV_FILE` in the hook's environment (e.g. a hostile project `settings.json` env block, if that propagates to hook processes — unverified) gets bash execution on the next event. Cheap hardening before sourcing: require the file to be owned by the current user with no group/other permissions (`[ -O "$ENV_FILE" ]` + mode 600/400 check), or drop the production override entirely and point the bats suite at a fake `$HOME` instead.

home/dot_config/ntfy/private_server.yml.tmpl
```diff
@@ -0,0 +1,32 @@
...
+
+# Public URL of this server as seen by subscribers (the `tailscale serve`
+# HTTPS endpoint). Stored in 1Password so the tailnet's MagicDNS name never
+# lands in this public repo.
+base-url: "{{ onepasswordRead "op://kryota.dev/Dotfiles - ntfy/base-url" | trim }}"
+
+# iOS instant push relay (user-approved at the #337 PRD gate): forwards poll
+# requests (topic metadata only, never message bodies) through ntfy.sh/APNs.
+# The sole approved metadata egress beyond the tailnet.
+upstream-base-url: "https://ntfy.sh"
```
kryota-dev, line 17: [SHOULD] Scope note: per docs.ntfy.sh/config, `upstream-base-url` publishes a poll request to ntfy.sh for **all incoming messages**, regardless of whether any iOS device subscribes — the "iOS instant push" framing in the comment and in notifications.md understates the egress (session-activity timing metadata flows to ntfy.sh on every publish). Fix the doc wording; as a follow-up, consider `upstream-access-token` so third parties cannot subscribe to the derived upstream topic, and if the smoke test shows iOS instant push does not work tailnet-only, remove this key to bring egress to zero (the PRD already allows that degrade).

home/run_once_after_11-validate-1password.sh.tmpl
```diff
@@ -36,6 +36,15 @@ ITEMS=(
...
   # exposing a `pattern` field holding a single `name1|name2|...` alternation.
   "op://kryota.dev/Dotfiles - Redact Patterns/pattern"
+  # Self-hosted ntfy notification server (#337). Create one item exposing:
+  #   base-url   — the `tailscale serve` HTTPS endpoint (https://<host>.<tailnet>.ts.net),
+  #                kept out of this public repo (rendered into ~/.config/ntfy/server.yml)
+  #   credential — the write-only publisher token from `ntfy token add`
+  #                (rendered into the 0600 ~/.config/ntfy/notify-env)
+  # The read-only subscriber token also lives in this item but is entered on
+  # devices manually from 1Password, so it is not validated here.
+  "op://kryota.dev/Dotfiles - ntfy/base-url"
```
kryota-dev, line 46: [SHOULD] The two new references get existence checks only, while the neighboring Redact Patterns item gets content validation. A `base-url` containing a quote/newline can escape the YAML scalar in server.yml (config-key injection, e.g. flipping `auth-default-access`), and a `credential` containing `'` breaks the `squote` wrapping sourced by the hook. Mirror the existing precedent: assert `https://*.ts.net` shape and no quote/newline for base-url, and a `tk_…` token shape with no single quote for credential.

home/.chezmoidata.toml
```diff
@@ -162,3 +162,18 @@ ghq_user = "kryota-dev"
...
+  # manager can track it (drift is caught by tests/files.bats).
+  port = 2586
+  # Server-side history retention (cache-duration). User-approved trade-off:
+  # 7 days of reviewable history vs. plaintext SQLite residency on disk.
+  cache_duration = "168h"
+  # Fixed urgency-based topics; attribution (repo/branch/account/session/event)
+  # travels in tags and titles, never in topic names.
+  topic_attention = "claude-attention"
+  topic_done = "claude-done"
+  topic_test = "claude-test"
```
kryota-dev: [SHOULD] `topic_test` is declared in the `[ntfy]` SSOT block but nothing references it — the templates only consume `topic_attention`/`topic_done`, and the docs hardcode `claude-test` as a literal. Either wire it through (`NTFY_TOPIC_TEST` in the env template, referenced by the smoke-test runbook) or drop the key; as-is the SSOT claim in the comment above is not true for this value.

docs/architecture/notifications.md
```diff
@@ -0,0 +1,130 @@
...
+   in a separate field (entered on devices manually), then re-run `chezmoi apply`.
+5. **Subscribe devices** — ntfy app (iOS/Android) → *Use another server* with the
+   `base-url` endpoint, the subscriber token from 1Password, and the two topics.
+
+## Smoke test
+
+```bash
+# anonymous publish must be denied (401/403)
+curl -s -o /dev/null -w '%{http_code}\n' -d test "$(op read 'op://kryota.dev/Dotfiles - ntfy/base-url')/claude-test"
+# read-only token must be denied for publish (403)
```
kryota-dev, line 114: [SHOULD] These smoke-test examples pass tokens via `-H "Authorization: Bearer <token>"` — argv-visible in `ps` and persisted to shell history, which is exactly the pattern the wrapper's own bats tests forbid for the production path. Rewrite the examples with `-K` process substitution feeding the header from `op read` so operators following the runbook do not leak what the code protects. Also note that `ntfy token add` echoes tokens into terminal scrollback.

home/dot_config/ntfy/compose.yaml.tmpl
kryota-dev: **Fixed in ae39275 ** — switched to the official `binwiederhier/ntfy` with an exact tag + digest pin (`v2.26.0@sha256:a6d33506…`). The Renovate regex manager (`matchStrings`/`depNameTemplate`) and both bats assertions now track the correct name, and a namespace assertion (`! grep binary/ntfy` + literal `binwiederhier` regex) prevents recurrence. Also added an explicit compose project name (`ntfy-dotfiles`) to scope `--remove-orphans`.

home/run_onchange_after_31-setup-ntfy.sh.tmpl
kryota-dev: **Fixed in ae39275 ** — (1) `docker compose up -d` is now wrapped in warn-and-skip so pull/config failures can no longer hard-fail `chezmoi apply` (a missing docker CLI gets its own distinct message too); (2) the recovery wording in the script comment, notifications.md/.ja troubleshooting, and the run_onchange caveat now states accurately that exit-0 records the state and only the printed `docker compose up -d` command recovers — re-running apply alone does not retry.

home/dot_claude/executable_ntfy-notify.sh
kryota-dev: **Fixed in ae39275 ** — order swapped to scrub-then-truncate with an 8 KiB byte pre-cap bounding the perl input, so a boundary-straddling identifier is redacted before the cut and the final message stays within 200 codepoints. Covered by a new bats case (195×a + `acmecorp` + tail ⇒ no `acme` remnant, length ≤ 200).

home/dot_claude/executable_ntfy-notify.sh
kryota-dev: **Fixed in ae39275 ** — `repo` and `branch` now pass through the client-identifier scrub before landing in the title and tags, closing the gap where Notification events published attribution unscrubbed.

home/dot_claude/executable_ntfy-notify.sh
kryota-dev: **Fixed in ae39275 ** — trap extended to `EXIT INT TERM HUP` (single-quoted, evaluated at fire time) so a timeout kill no longer strands the token-bearing curl config in TMPDIR.

home/dot_claude/executable_ntfy-notify.sh
kryota-dev: **Fixed in ae39275 ** — before sourcing, the wrapper now requires the env file to be owned by the current user with mode 0600/0400 (BSD/GNU stat both handled); anything else fails open without publishing. Covered by a new group-readable-env bats case; the fixture env files are chmod 600 accordingly.

home/dot_config/ntfy/private_server.yml.tmpl
kryota-dev: **Fixed in ae39275 ** — the config comment and notifications.md/.ja now state accurately that a poll request goes to ntfy.sh for every incoming message regardless of iOS subscriptions, with the egress-zero removal path documented for a failed smoke test. `upstream-access-token` stays a follow-up (needs a new 1Password field; PRD keeps it out of v1 scope).

home/run_once_after_11-validate-1password.sh.tmpl
kryota-dev: **Fixed in ae39275 ** — the gate now content-validates both fields mirroring the Redact Patterns precedent: base-url must be an `https://…ts.net` endpoint with no quote/newline (YAML-scalar injection guard), credential must be a `tk_…` token with no single quote (squote guard).

home/.chezmoidata.toml
kryota-dev: **Fixed in ae39275 ** — dropped the dead `topic_test` key; the block comment now states explicitly that the smoke-test topic is a runbook-only convention in notifications.md, not rendered config, so the SSOT claim matches reality.

docs/architecture/notifications.md
kryota-dev: **Fixed in ae39275 ** — smoke-test and history examples rewritten to feed tokens via `curl -K` process substitution (`ro()`/`wo()` helpers reading from 1Password), consistent with the rule the wrapper enforces; added the `ntfy token add` scrollback-clearing note. JA mirror synced.

PR #354, opened 2026-07-26
Consolidate the sdd review into a single post-PR review pass

kryota-dev:
## Summary

Consolidates the two review passes that ran for the same change on the `/pr-workflow` standard/large path into a single post-PR review pass, per #347.

Previously `/sdd` ran a pre-commit built-in review (a dynamic reviewer team that looped until APPROVE, leaving no trace on GitHub) **and** `/pr-workflow` ran `/multi-review` on the created PR. The two overlapped on the same diff (code-quality/security duplicated), and sdd's findings could not be deduplicated by multi-review because they never reached GitHub.

Now review happens once, via the `pr-workflow` Phase 5-7 pipeline (`monitor-ci` → `multi-review` → `review-resolve-loop`), which already provides fix enforcement, GitHub records, and independent second opinions.

## Changes

- **`sdd`**: remove Phase 5 (built-in review team); renumber Phase 6/7 → 5/6; make it a fully internal component (drop `user-invocable`, add `disable-model-invocation: true`, strip the standalone description triggers). `sdd` now ends at PR creation.
- **`multi-review`**: add a cross-cutting specialist roster (`performance` / `test` / `ux`, keyed on change characteristics) and an opt-in `--spec-context <dir>` input so a caller can pass the spec docs for spec-implementation consistency checks. Existing behaviour is preserved when the flag is absent.
- **New reviewer agents**: `performance-reviewer` / `test-reviewer` / `ux-reviewer`, ported from the removed sdd Phase 5 role templates, following the existing language-specialist agent pattern (SSOT in the agent definition, `sonnet` + `high` pinned, read-only Bash, coverage-first).
- **`pr-workflow`**: replace the "dual review (併用＝役割分離)" decision with the single-pass design referencing this issue; pass `--spec-context` (absolute path) to `/multi-review` on the standard/large path.
- **Drive-by consistency fix**: correct stale `pr-workflow Phase 5` → `Phase 6` adversarial-verify references across the reviewer roster (the actual adversarial-verify step lives in pr-workflow Phase 6). This keeps the roster consistent after adding the three new agents.
- **Tests**: `tests/files.bats` gains coverage for the new agents' frontmatter, the multi-review roster/spec-context wiring, and sdd's review-phase removal (no review phase, renumbered phases, not user-invocable, model-invocation disabled).

## Acceptance criteria

- **AC-1**: `/pr-workflow` standard/large runs exactly one review pass (the post-PR `multi-review`); `sdd` no longer contains a review phase. ✅
- **AC-2**: `multi-review` can spawn performance/test/ux specialists and accepts spec-document context from the caller, so no review perspective is lost relative to the removed sdd Phase 5. ✅

## Out of scope

- Duplicate `/model-fitness-check` invocations across the pr-workflow → sdd → multi-review chain (deliberately deferred in #347).

## Testing

- `make test-bats` — 133/133 pass (including the 3 new assertions).
- `make lint` — shellcheck + shfmt + zsh syntax all pass.

Closes #347


kryota-dev (commented):
## Multi-Review integrated result — PR #354

Reviewers on the post-PR pipeline: **cc-code-review**, **cc-security-review**, **codex**. No existing reviews to dedup against.

### Verdict

|---|---|---|---|
| MUST / High | 0 | 1 (High) | 0 |
| SHOULD / Medium | 3 | 3 | 2 (P2) |
| NITS / Low | 3 | 1 | 0 |
| GOOD | 5 | 7 (noted) | — |

Security overall: **Medium**. The sole High is a **pre-existing, self-documented architecture tradeoff** — reviewer agents use unrestricted `Bash` gated only by a prompt-level read-only constraint (`tests/files.bats` documents this as the accepted design); this PR extends the same pattern to 3 new agents rather than introducing it.

### Findings addressed in this PR (commit `1eabb48`)

1. **[convergent across all 3 tools] sdd was not truly internal.** `user-invocable` defaults to `true` when absent, and `disable-model-invocation` only blocks *model* auto-invocation (verified against the Claude Code frontmatter docs). So `/sdd` was still user-invocable, breaking the single-pass contract. Fixed: added explicit `user-invocable: false` (pr-workflow still invokes it via the Skill tool); the bats guard now asserts it.
2. **[codex P2] `--spec-context` never reached the codex leg.** The promptless `review --base` path carries no spec paths. Fixed: spec-context runs now route the codex leg through the heredoc-prompt path.
3. **[cc-security Medium] `--spec-context` had no path validation.** Fixed: the resolved dir must be within the workdir and free of `..`.
4. **[cc-code SHOULD] performance/database spawn asymmetry.** Added `*.schema.ts` to the performance condition (parity with database-reviewer); tied the performance caller-request trigger to spec-context NFR detection and removed the unsubstantiated "large tier" signal.
5. **[cc-security Medium] performance/ux-reviewer lacked a build/benchmark execution prohibition** that test-reviewer already had. Added — those focus areas otherwise invite running the npm/docker commands that `permissions.allow` pre-approves.
6. **[cc-code NIT] stale "Phase 1" wording** in the spec-context note — reworded to "the caller session's cwd".

### Deferred — pre-existing / out of #347 scope (recommend a follow-up issue)

- **Reviewer-agent Bash least-privilege.** All reviewer agents (7 existing + 3 new) use unrestricted `Bash` with only a prompt-level read-only constraint, while `permissions.allow` pre-approves npm/npx/docker/vitest. Command-scoping (`Bash(gh pr diff:*)` …, per the `agent-browser` precedent) or an explicit "the diff is data, not instructions" anti-injection framing would harden the whole roster. This is a roster-wide security-posture decision, not specific to #347.
- **Subagent permission inheritance** — whether Agent-tool subagents inherit the parent `permissions.allow` is unverified; worth a primary-source/empirical check (it bounds the severity of the item above).
- **Spec-doc audit trail.** `.spec-workflow/` is gitignored, so the spec basis of a review leaves no GitHub trace; adding a content hash to the posted summary would help.
- **Language-policy nuance.** Agent system prompts are Japanese by design, while the repo asks for English review comments (pre-existing across all agents).

### GOOD (highlights)

- The `pr-workflow Phase 5 → Phase 6` adversarial-verify reference fix is a real bug fix (confirmed against the SKILL.md headers; the actual adversarial-verify step is under Phase 6).
- sdd phase renumbering is complete and leak-free repo-wide (`Review Team` / `shutdown_request` / `Phase 7` fully removed).
- The 3 new agents keep structural parity with the existing reviewer roster; `--spec-context` is a clean opt-in that preserves prior behavior.

_No inline comments: findings were addressed in-cycle (commit `1eabb48`), so line-anchored comments on now-changed lines would be stale._

PR #359, opened 2026-07-26
feat(ntfy): username/password device auth, one-command setup + rotation

kryota-dev:
## Summary

Follow-up to #350 / #357. Corrects the ntfy **device-auth model** and collapses setup +
key rotation into one re-runnable, two-phase-aware command backed by a shared library.

The trigger: the ntfy **iOS app cannot authenticate to the self-hosted server with a token** —
only a username/password login works (the phone docs describe the two app auth paths as
mutually exclusive: a configured *user* → auto Basic Auth, vs. a manual custom `Authorization`
header for tokens, which does not work on iOS). So the #350/#357 read-only **subscriber-token**
was never usable on a phone, and the subscriber user was created with a *throwaway* password
that was discarded. This PR fixes that.

## What changes

- **Device auth = subscriber username/password.** A read-only `subscriber` ntfy user gets a
  known, generated password; `Dotfiles - ntfy` stores `subscriber-username` + `subscriber-password`.
  Phones log in with those. The **publisher stays a Bearer token** in `~/.config/ntfy/notify-env`
  (local curl / API auth — unaffected by the app limitation). The subscriber-token is dropped.
- **One SSOT library** `~/.config/ntfy/lib.sh` (`home/dot_config/ntfy/lib.sh.tmpl`) drives the
  whole lifecycle. It is sourced by `run_onchange_after_31-setup-ntfy` (apply-time, **fail-open**)
  and by `~/.local/bin/ntfy-setup` (user command, **fail-clear**). `ntfy-setup` runs the full
  flow — `docker compose up -d` → `tailscale serve --bg` → provision/repair — and `--rotate
  publisher|subscriber|all`. It exits non-zero with guidance only when a prerequisite it cannot
  fix (Docker Desktop, `tailscale up`) is missing. This is also what restarts a container that
  went down after a reboot, which `chezmoi apply` alone cannot (run_onchange records exit-0).
- **`base-url` dropped everywhere** — it is optional for our use (only attachments / email need
  it, unused here; it does not affect the iOS `upstream-base-url` relay). Removed from
  `server.yml`, from the 1Password validate gate (both the ITEMS entry **and** the format-check
  stanza), so nothing ntfy blocks `chezmoi apply`. The MagicDNS URL is derived on demand only.
- **Two-phase aware.** At first setup Tailscale is not `up` and Docker is not running, so
  provisioning cannot finish during the first apply. `install.sh` notes this; `ntfy-setup` is
  the single command to run once both are available. The 1Password item is auto-created only
  when absent (never overwritten).
- **Rotation is correct.** Publisher: read the old token first → issue new → write notify-env →
  `ntfy token remove` the old one. Subscriber: `NTFY_PASSWORD=… ntfy user change-pass` (preserves
  ACLs) + update 1Password. Docs corrected: `ntfy token add` is additive, not invalidating.

## Backward compatibility

Non-destructive. The library never overwrites an existing item, so legacy `base-url` /
`credential` / `subscriber-token` fields are left harmless; `server.yml` re-renders without
base-url; `notify-env` is unchanged. Running `ntfy-setup` on this machine **repairs** the
subscriber user (throwaway → known password) and records username/password in 1Password.

## Verification

- `chezmoi execute-template` render + `bash -n` + `shellcheck` for the lib, script 31, server.yml,
  validate; `ntfy-setup` `bash -n` + `shellcheck`. Rendered notify-env / password writes carry
  the secret only through the heredoc / `NTFY_PASSWORD` env — never argv (except the documented
  `op item edit` trade-off), never `set -x`, never echoed.
- `make test` green (lint + 232 bats). FACT markers resynced (`onepassword-vault-item-count` 4;
  `ci-both-exclusion-count` 7) and a pre-existing `secrets-1password.md` CI-exclusion drift
  (listed 6, missing `server.yml`) fixed to 7 with a FACT marker.
- No tailnet MagicDNS name or client identifier literal in the repo.

## Design record

`.claude/prds/337-ntfy-setup-rotation.prd.md` (grill-me `--mode=auto`, adversarially verified).

## Post-merge (your environment)

Run `ntfy-setup` on this Mac (after `tailscale up` + Docker running) to repair the subscriber
login and store its username/password; then log the phone into the server with username
`subscriber` and the password from 1Password. Legacy `base-url` / `credential` /
`subscriber-token` fields on the item may be deleted manually (harmless if left).

---

_Process note: this large-tier change was run on Opus 4.8, below the Opus 5 floor the workflow
requests for large tier (per an explicit continue-anyway); the design deliberation / adversarial
verification may be below the Opus 5 baseline._

### DIFF
--- a/home/dot_config/ntfy/lib.sh.tmpl
+++ b/home/dot_config/ntfy/lib.sh.tmpl
@@ -9,9 +9,11 @@
 #   - This file is SOURCED. It never calls `exit`; functions `return` a status
 #     and the caller decides the exit policy. It never enables `set -x`.
 #   - No secret is ever echoed or traced. Passwords reach ntfy via the
-#     NTFY_PASSWORD env (never argv); the only argv exposure is the local
-#     `op item edit field=value` call, the same transient trade-off already
-#     accepted for tokens (op has no stdin path for single-field edits).
+#     NTFY_PASSWORD env (never argv). Two local, transient argv exposures are
+#     accepted: `op item edit field=value` (op has no single-field stdin
+#     shortcut — only a full JSON template), and the about-to-be-revoked old
+#     token passed to `ntfy token remove` (it takes the token positionally,
+#     with no env alternative). Both are local, one-shot, own-machine calls.
 #   - Template values live on assignment-only lines (the lint pipeline strips
 #     every line containing a template action before shellcheck); references use
 #     the ${var:-} form so the stripped body still lints clean. No comment or
@@ -74,7 +76,10 @@ ntfy_magicdns() {
   if command -v jq >/dev/null 2>&1; then
     dns="$(printf '%s' "$json" | jq -r '.Self.DNSName // empty' 2>/dev/null || true)"
   else
-    dns="$(printf '%s' "$json" | grep -oE '"DNSName"[[:space:]]*:[[:space:]]*"[^"]*"' | head -n 1 | sed 's/.*"\(.*\)"/\1/')"
+    # Scope to the .Self object before grabbing DNSName — JSON key order is not
+    # guaranteed, so a bare "first DNSName" could be a peer's. "Self" appears
+    # once at top level; strip everything up to it, then take the next DNSName.
+    dns="$(printf '%s' "$json" | sed 's/.*"Self"[[:space:]]*:[[:space:]]*{//' | grep -oE '"DNSName"[[:space:]]*:[[:space:]]*"[^"]*"' | head -n 1 | sed 's/.*"\(.*\)"/\1/')"
   fi
   dns="${dns%.}"
   [ -n "$dns" ] || return 1
@@ -119,7 +124,10 @@ ntfy_wait_ready() {
 # --- User / ACL / token primitives ------------------------------------------
 
 ntfy_user_exists() {
-  ntfy_exec user list 2>/dev/null | grep -q "user ${1}"
+  # Match the name exactly: the next char after it must be a field delimiter
+  # (space / colon / paren) or end of line, so `publisher` never matches a
+  # hypothetical `publisher-foo`. Does not assume where "user " sits on the line.
+  ntfy_exec user list 2>/dev/null | grep -qE "user ${1}([ :(]|$)"
 }
 
 # Create a user with a throwaway random password if it does not exist. Used for
@@ -190,9 +198,10 @@ ntfy_read_notify_env_token() {
 }
 
 ntfy_notify_env_has_token() {
+  # Only real server-issued tokens (tk_…) are ever written here; there is no
+  # placeholder-seeding path, so a shape check is sufficient.
   [ -f "$ntfy_env_file" ] &&
-    grep -qE '^NTFY_TOKEN=.*tk_' "$ntfy_env_file" &&
-    ! grep -qE '^NTFY_TOKEN=.*tk_REPLACE' "$ntfy_env_file"
+    grep -qE '^NTFY_TOKEN=.*tk_' "$ntfy_env_file"
 }
 
 # --- 1Password item (subscriber username/password) ---------------------------
@@ -211,8 +220,11 @@ ntfy_ensure_item() {
     "subscriber-password[password]=${ntfy_placeholder}" >/dev/null
 }
 
+# Prints the stored password (empty if the field is absent/empty) and — crucially
+# — propagates op's exit status so callers can tell a real op failure from a
+# genuinely empty value. Callers that overwrite on "empty" MUST check the status.
 ntfy_read_subscriber_password() {
-  op read "op://${ntfy_op_vault}/${ntfy_op_item}/subscriber-password" 2>/dev/null || true
+  op read "op://${ntfy_op_vault}/${ntfy_op_item}/subscriber-password" 2>/dev/null
 }
 
 ntfy_store_subscriber_password() {
@@ -224,12 +236,19 @@ ntfy_store_subscriber_password() {
 # --- High-level provisioning -------------------------------------------------
 
 ntfy_provision_publisher() {
+  local existed=1
+  ntfy_user_exists "$ntfy_pub_user" || existed=0
   ntfy_ensure_user "$ntfy_pub_user" || {
     ntfy_warn "could not create the publisher user; provision manually."
     return 1
   }
   ntfy_grant_acl "$ntfy_pub_user" write-only
-  if ntfy_notify_env_has_token; then
+  # Only trust an existing notify-env token if the publisher user was already
+  # present. If we just (re)created it — e.g. after a user.db loss — any token in
+  # notify-env is stale against the rebuilt server, so reissue. This mirrors the
+  # subscriber self-heal and makes "delete user.db, run ntfy-setup" actually
+  # restore the publisher too.
+  if [ "$existed" = 1 ] && ntfy_notify_env_has_token; then
     return 0
   fi
   local tok
@@ -254,8 +273,12 @@ ntfy_provision_subscriber() {
     ntfy_warn "could not ensure the 1Password item; skipping subscriber."
     return 1
   }
-  local stored pw
-  stored="$(ntfy_read_subscriber_password)"
+  local stored pw rc=0
+  stored="$(ntfy_read_subscriber_password)" || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    ntfy_warn "could not read the stored subscriber password (op error); leaving the existing subscriber credential untouched to avoid a device lockout."
+    return 1
+  fi
   if [ -n "$stored" ] && ! printf '%s' "$stored" | grep -q "$ntfy_placeholder"; then
     ntfy_ensure_user_with_password "$ntfy_sub_user" "$stored" || {
       ntfy_warn "could not reconcile the subscriber user password."
@@ -306,8 +329,13 @@ ntfy_rotate_publisher() {
   }
   ntfy_write_notify_env "$new" || return 1
   if [ -n "$old" ] && [ "$old" != "$new" ]; then
-    ntfy_exec token remove "$ntfy_pub_user" "$old" >/dev/null 2>&1 ||
-      ntfy_warn "could not revoke the old publisher token (it may already be gone)."
+    if ! ntfy_exec token remove "$ntfy_pub_user" "$old" >/dev/null 2>&1; then
+      # The new token is already active, but the OLD one may still be valid —
+      # for a leaked-token rotation that is the whole point, so surface it
+      # (non-zero) instead of reporting success.
+      ntfy_warn "the new publisher token is active, but revoking the old one failed; it may still be valid. Retry 'ntfy-setup --rotate publisher' or remove it manually."
+      return 1
+    fi
   fi
   ntfy_log "publisher token rotated."
 }
@@ -333,7 +361,9 @@ ntfy_creds_exist() {
   ntfy_notify_env_has_token || return 1
   command -v op >/dev/null 2>&1 || return 1
   local stored
-  stored="$(ntfy_read_subscriber_password)"
+  # An op read error here is conservatively treated as "no usable creds" (the
+  # interactive path then provisions rather than prompting to rotate).
+  stored="$(ntfy_read_subscriber_password || true)"
   [ -n "$stored" ] && ! printf '%s' "$stored" | grep -q "$ntfy_placeholder"
 }
 
@@ -345,8 +375,8 @@ ntfy_prompt_yn() {
     read -r ans </dev/tty || ans=""
   fi
   case "$ans" in
-  [yY] | [yY][eE][sS]) return 0 ;;
-  *) return 1 ;;
+    [yY] | [yY][eE][sS]) return 0 ;;
+    *) return 1 ;;
   esac
 }
 
--- a/home/dot_config/ntfy/lib.sh.tmpl
+++ b/home/dot_config/ntfy/lib.sh.tmpl
@@ -1,0 +1,365 @@
+# shellcheck shell=bash
+# ntfy lifecycle library (kryota-dev/dotfiles#337). Deployed 0644 to
+# ~/.config/ntfy/lib.sh and sourced by run_onchange_after_31-setup-ntfy (apply
+# time, fail-open) and ~/.local/bin/ntfy-setup (user command, fail-clear). This
+# is the single source of truth for bringing the server up and provisioning /
+# rotating credentials, so the two entry points can never drift.
+#
+# Design contract:
+#   - This file is SOURCED. It never calls `exit`; functions `return` a status
+#     and the caller decides the exit policy. It never enables `set -x`.
+#   - No secret is ever echoed or traced. Passwords reach ntfy via the
+#     NTFY_PASSWORD env (never argv); the only argv exposure is the local
+#     `op item edit field=value` call, the same transient trade-off already
+#     accepted for tokens (op has no stdin path for single-field edits).
+#   - Template values live on assignment-only lines (the lint pipeline strips
+#     every line containing a template action before shellcheck); references use
+#     the ${var:-} form so the stripped body still lints clean. No comment or
+#     control-flow line contains a literal open-brace pair.
+ntfy_port="{{ .ntfy.port }}"
+ntfy_topic_attention="{{ .ntfy.topic_attention }}"
+ntfy_topic_done="{{ .ntfy.topic_done }}"
+
+ntfy_compose_cfg="$HOME/.config/ntfy/compose.yaml"
+ntfy_env_file="$HOME/.config/ntfy/notify-env"
+ntfy_state_dir="$HOME/Library/Application Support/ntfy"
+ntfy_serve_target="http://127.0.0.1:${ntfy_port:-}"
+ntfy_op_item="Dotfiles - ntfy"
+ntfy_op_vault="kryota.dev"
+ntfy_smoke_topic="claude-test"
+ntfy_pub_user="publisher"
+ntfy_sub_user="subscriber"
+# Placeholder marker the 1Password bootstrap password carries until provisioned.
+ntfy_placeholder="REPLACE_ME"
+
+ntfy_log() { echo "[ntfy] $*"; }
+ntfy_warn() { echo "[ntfy] $*" >&2; }
+
+ntfy_ensure_state_dir() {
+  mkdir -p "$ntfy_state_dir"
+  chmod 700 "$ntfy_state_dir"
+}
+
+# --- Prerequisite probes -----------------------------------------------------
+
+ntfy_docker_ready() {
+  command -v docker >/dev/null 2>&1 && docker info >/dev/null 2>&1
+}
+
+ntfy_tailscale_bin() {
+  local b
+  b="$(command -v tailscale 2>/dev/null || true)"
+  if [ -z "$b" ] && [ -x "/Applications/Tailscale.app/Contents/MacOS/Tailscale" ]; then
+    b="/Applications/Tailscale.app/Contents/MacOS/Tailscale"
+  fi
+  printf '%s' "$b"
+}
+
+ntfy_tailscale_up() {
+  local b
+  b="$(ntfy_tailscale_bin)"
+  [ -n "$b" ] || return 1
+  # BackendState is "Running" once tailscaled is up and logged in.
+  "$b" status --json 2>/dev/null | grep -q '"BackendState"[[:space:]]*:[[:space:]]*"Running"'
+}
+
+# MagicDNS name of this host (for human-facing device-login / smoke-test hints
+# only; never stored). Trailing dot stripped. Empty if unavailable.
+ntfy_magicdns() {
+  local b json dns
+  b="$(ntfy_tailscale_bin)"
+  [ -n "$b" ] || return 1
+  json="$("$b" status --json 2>/dev/null || true)"
+  [ -n "$json" ] || return 1
+  if command -v jq >/dev/null 2>&1; then
+    dns="$(printf '%s' "$json" | jq -r '.Self.DNSName // empty' 2>/dev/null || true)"
+  else
+    dns="$(printf '%s' "$json" | grep -oE '"DNSName"[[:space:]]*:[[:space:]]*"[^"]*"' | head -n 1 | sed 's/.*"\(.*\)"/\1/')"
+  fi
+  dns="${dns%.}"
+  [ -n "$dns" ] || return 1
+  printf '%s' "$dns"
+}
+
+# --- Server bring-up ---------------------------------------------------------
+
+ntfy_start_server() {
+  ntfy_ensure_state_dir
+  docker compose -f "$ntfy_compose_cfg" up -d --remove-orphans
+}
+
+# Assert the tailnet HTTPS front. `--bg` persists across reboots. NEVER use
+# `tailscale funnel` here: the server must stay tailnet-only.
+ntfy_assert_serve() {
+  local b
+  b="$(ntfy_tailscale_bin)"
+  [ -n "$b" ] || {
+    ntfy_warn "tailscale CLI not found; cannot assert serve mapping."
+    return 1
+  }
+  "$b" serve --bg "${ntfy_serve_target:-}"
+}
+
+ntfy_exec() {
+  docker compose -f "$ntfy_compose_cfg" exec -T ntfy ntfy "$@"
+}
+
+# Give the freshly-started container a moment to accept `ntfy user` calls.
+ntfy_wait_ready() {
+  local i
+  for i in 1 2 3 4 5 6 7 8; do
+    if ntfy_exec user list >/dev/null 2>&1; then
+      return 0
+    fi
+    sleep 1
+  done
+  return 1
+}
+
+# --- User / ACL / token primitives ------------------------------------------
+
+ntfy_user_exists() {
+  ntfy_exec user list 2>/dev/null | grep -q "user ${1}"
+}
+
+# Create a user with a throwaway random password if it does not exist. Used for
+# the publisher, which authenticates with a token, so its password is never
+# needed or stored.
+ntfy_ensure_user() {
+  local u="$1"
+  ntfy_user_exists "$u" && return 0
+  ntfy_log "creating ntfy user ${u}..."
+  NTFY_PASSWORD="$(openssl rand -base64 24)" \
+    docker compose -f "$ntfy_compose_cfg" exec -T -e NTFY_PASSWORD ntfy ntfy user add "$u"
+}
+
+# Ensure a user exists with a specific (known) password. `user add` when absent,
+# `user change-pass` when present — both read the password from NTFY_PASSWORD
+# (verified in the ntfy CLI source), so the password never lands in argv, and
+# change-pass preserves the user's ACLs.
+ntfy_ensure_user_with_password() {
+  local u="$1" pw="$2" sub
+  if ntfy_user_exists "$u"; then
+    sub="change-pass"
+  else
+    sub="add"
+  fi
+  NTFY_PASSWORD="$pw" \
+    docker compose -f "$ntfy_compose_cfg" exec -T -e NTFY_PASSWORD ntfy ntfy user "$sub" "$u"
+}
+
+# Idempotent ACL grant across the two rendered topics plus the runbook-only
+# smoke-test topic.
+ntfy_grant_acl() {
+  local u="$1" perm="$2" t
+  for t in "${ntfy_topic_attention:-}" "${ntfy_topic_done:-}" "$ntfy_smoke_topic"; do
+    ntfy_exec access "$u" "$t" "$perm" >/dev/null 2>&1 || true
+  done
+}
+
+ntfy_issue_token() {
+  ntfy_exec token add --label "$1" "$2" 2>/dev/null | grep -oE 'tk_[A-Za-z0-9]+' | head -n 1
+}
+
+# --- notify-env (publisher token, runtime state) -----------------------------
+
+# Write notify-env 0600 via umask; the token reaches the file only through the
+# heredoc (never echoed or traced). Reuses the exact format the wrapper sources.
+ntfy_write_notify_env() {
+  (
+    umask 077
+    cat >"$ntfy_env_file" <<EOF
+# ntfy publisher credentials for the Claude Code hook wrapper
+# (~/.claude/ntfy-notify.sh, kryota-dev/dotfiles#337). Runtime state written
+# 0600 by ~/.config/ntfy/lib.sh — NOT a chezmoi target. Sourced, never exported,
+# by the wrapper so the token only lives in that process. The wrapper publishes
+# to the loopback listener directly, so it keeps working even if the tailscale
+# serve front is down.
+NTFY_URL='${ntfy_serve_target}'
+# Write-only publisher token (least privilege: cannot read history if leaked).
+NTFY_TOKEN='$1'
+NTFY_TOPIC_ATTENTION='${ntfy_topic_attention}'
+NTFY_TOPIC_DONE='${ntfy_topic_done}'
+EOF
+  )
+}
+
+ntfy_read_notify_env_token() {
+  [ -f "$ntfy_env_file" ] || return 1
+  sed -n "s/^NTFY_TOKEN='\{0,1\}\(tk_[A-Za-z0-9]*\).*/\1/p" "$ntfy_env_file" | head -n 1
+}
+
+ntfy_notify_env_has_token() {
+  [ -f "$ntfy_env_file" ] &&
+    grep -qE '^NTFY_TOKEN=.*tk_' "$ntfy_env_file" &&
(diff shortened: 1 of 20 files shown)


home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+  local old new
+  old="$(ntfy_read_notify_env_token || true)"
+  new="$(ntfy_issue_token chezmoi "$ntfy_pub_user" || true)"
+  [ -n "$new" ] || {
+    ntfy_warn "new publisher token issuance failed."
+    return 1
+  }
+  ntfy_write_notify_env "$new" || return 1
+  if [ -n "$old" ] && [ "$old" != "$new" ]; then
+    ntfy_exec token remove "$ntfy_pub_user" "$old" >/dev/null 2>&1 ||
```
kryota-dev: [SHOULD] Rotation reports success even when revocation fails. If `ntfy token remove` errors, this warns but `ntfy_rotate_publisher` still falls through to the success log and returns 0 — so `ntfy-setup --rotate publisher` prints "rotated" and exits 0 while the old (possibly leaked) token can stay valid, defeating the leaked-token use case on a transient Docker/ntfy hiccup. Distinguish "already gone" (idempotent re-run — fine) from a real failure and surface the latter so the operator retries revocation.

Separately: `$old` (a real `tk_…` bearer credential) lands on argv here (`docker compose exec … ntfy token remove publisher tk_OLD`), momentarily `ps`-observable. `ntfy token remove` has no env path, so this is unavoidable in impl — but the stated hygiene invariant ("argv exposure is only `op item edit`", requirements.md / AC-006) should be reworded to name this as a second accepted local/transient exposure.

kryota-dev (commented):
## Multi-Review 統合レビュー結果 — PR #359

5 レビュー（cc-code-review / cc-security-review @opus / test-reviewer / architecture-reviewer / codex）を並列実行し、各指摘を lib.sh 実体・spec で一次検証しました。既存レビューは 0 件（重複除外なし）。

### 総合評価

| カテゴリ | 件数（裁定後） |
|---|---|
| MUST（コード correctness） | **0** |
| SHOULD | 8 |
| NITS | 4 |
| GOOD | 多数（下記） |

- **セキュリティ総合リスク: Low**（Critical/High 0）。リモート悪用経路なし: loopback バインド + tailnet 限定 serve（funnel 不使用を bats で pin）、`auth-default-access: deny-all`、publisher=write-only / subscriber=read-only、secret は原則 env 経由。
- **severity 裁定**: test-reviewer が MUST とした 2 件（R-008 対話フロー未テスト・#357 parse 回帰の再現不能）はいずれも「コードは正しく動くがテスト未整備」= カバレッジ欠落のため **SHOULD へ再分類**。実バグ MUST は 0。large tier adversarial round: MUST 0 のため反証対象なし。昇格 SHOULD は各々 lib.sh 実体で直接検証済み。

### [SHOULD] 修正推奨（インラインで個別投稿）

1. **rotate_publisher: 旧トークン失効失敗でも success を返す + 旧トークンが argv 露出**（lib.sh.tmpl 付近）— codex + security の収束。漏洩トークンのローテーション目的で「失効完了」と誤報告しうる。「既に消えている（冪等再実行）」と「一時失敗」を区別する対応、および argv 露出（`ntfy token remove` は env 経路なし）を「2 つ目の許容 transient trade-off」として文書化。
2. **user.db 消失後に publisher token が再発行されない**（lib.sh.tmpl）— `ntfy_notify_env_has_token` は形式のみ検査。DB 再構築後も旧トークンが形式を通過し再発行されず、hook が死んだトークンで publish しつつ success 誤報告。subscriber 半分は self-heal するのに publisher 半分は非対称。publisher user を再作成した場合は再発行する。
3. **op read 失敗 = 空の誤同一視で実 subscriber-password を上書き**（lib.sh.tmpl）— transient な op 失敗（生体認証再要求/セッションタイムアウト）で新パスワード生成→上書き→全デバイスがサイレントにログイン不能。op read の exit code を捕捉し、エラー時は上書きせず温存。
4. **オーケストレーション順序が 2 エントリポイントに重複**（architecture）— bring-up 列（docker_ready→start_server→assert_serve）が run_onchange_after_31 と ntfy-setup に別々にベタ書き。「SSOT で never drift」目標が部分達成。`ntfy_bring_up` を lib に集約し両呼び出し元は 1 関数を呼ぶ形へ。
5. **R-008 対話ローテーション確認フローが未テスト**（test）— `ntfy_creds_exist` / `ntfy_prompt_yn` / `/dev/tty`（既存クレデンシャルあり・非 tty で既定 N）が bats で一切 pin されていない。本 PR の中核安全設計の退行を検出できない。
6. **#357 parse 回帰を再現できるテストが無い**（test）— `sed '/{{/d' | bash -n` は該当行を除去してから検査するため、コメント内リテラル `{{` の Go template パースエラーを検出できない。requirements.md は「#357 regression must not recur」を明示。`chezmoi execute-template --source` の実レンダリング assertion を追加（claude-plugin-reconciler テストに前例あり）。
7. **base-url format 検証 stanza 削除（R-003）が未テスト**（test）— stanza が将来復活すると op read が空を返し新規マシンの apply を毎回ハードフェイルさせる。validate-1password に不在 assertion を追加。
8. **install.sh の two-phase 案内（R-009）が未テスト**（test）— bootstrap テストは存在確認のみ。Darwin 限定 ntfy-setup 案内 + 「bootstrap では provisioning しない」を pin。

### [NITS]

- **funnel grep の不整合**: files.bats:1919 の `grep -v '#'`（行内に # があれば除外）は 1965 の anchored `^[^#]*...` より緩い。インラインコメント末尾の funnel 使用を見逃す。anchored 方式へ統一。
- **magicdns jq-less フォールバックが peer の DNSName を拾い得る**（lib.sh.tmpl:77）— `.Self` が `.Peer[]` より先という JSON キー順は非保証。fallback を `.Self` に限定するか jq 不在時は URL を伏せる。表示は人間向けヒントのみで実害は contrived。
- **ntfy_user_exists の部分一致 grep**（lib.sh.tmpl:122）— `user publisher` が `user publisher-foo` にも一致。ユーザー名は固定定数で悪用不可だが `grep -qw`/anchor が堅牢。
- **テスト名 "is on PATH" の overclaim**（files.bats:1971）— 本文は PATH を検証していない。改名するか PATH assertion を追加。

### [GOOD]

- `op item create` の non-destructive guard（`op item get` 存在確認 → 不在時のみ create、既存フィールド温存）。repo 初の 1Password 書き込みを安全に導入。
- secret hygiene: notify-env は umask 077 + heredoc で 0600、token/password は env・heredoc 経由のみで echo/trace/set -x 不使用。password は NTFY_PASSWORD env（`add`/`change-pass` 双方）。
- fail-open（script 31 は全経路 exit 0、bats で `exit 1` 不在を pin）と fail-clear（ntfy-setup は EXIT_ERR/USAGE/PREREQ で prereq 不能を明示）の一貫した分離。lib は `exit` を呼ばず return のみ。
- base-url を 4 箇所（server.yml / 1Password / ITEMS / format 検証 stanza）から除去、MagicDNS はオンデマンド導出のみ。`*.ts.net` リテラルの混入なし。
- 退行防止ガード（`credential[concealed]` / `subscriber-token` / funnel の再導入禁止）と ITEMS の `op://.../ntfy/` プレフィックス一般化。
- 端末認証を subscriber username/password に是正（#350/#357 の iOS で使えない subscriber-token を修正）、publisher token は API 認証専用に分離。read-old→issue→write→remove-old のローテーション順序が正しく実装（失効前に必ず新トークンを書き込む fail-safe）。

### ドキュメント follow-up（非ブロッカー）

- `docs/explanation/secrets-and-isolation.md` の全体設計図が repo 初の「1Password 書き込みパス」（`ntfy_ensure_item` / `op item edit`）を未反映（読み取り専用フローのみ記述）。恒久ドキュメントに新しいセキュリティ境界を追記推奨。
- `docs/contributing/ci-and-tests.md:127` の「1Password-backed secret template」という除外リストの framing は、base-url 除去後の server.yml（もう op-backed でない）に対して不正確。件数 FACT（7）と server.yml 掲載自体は正しい。AC-011 が rewording を宣言していた軽微な precision gap。

### 検証・裁定ログ

- codex [P1/P2]・security [Low×2] は lib.sh 実体（309/232/214-215 行）で制御フローを直接確認し採用。
- security の一部行番号参照（:1065 等）はファイル実体（366 行）と不一致だったが、関数名は正しく実体で再アンカー済み。
- arch [SHOULD]#2（ci-and-tests 未更新）は「件数・掲載は既に正しく、残るのは line 127 の framing 精度」と縮小して採用（NITS 相当の doc follow-up）。
- test-reviewer の MUST×2 は実体 files.bats を直接 grep 検証（1943-1987 に対話フロー/render assertion が不在）し SHOULD として採用。
- security [Info]（upstream-base-url egress・rotation 1 トークンのみ失効）は #350 承認済み by-design のため本 PR 対象外として棄却。

_pr-workflow large tier / multi-review --arch, spec-context 整合チェック込み。AI クレジット・署名なし。_

home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+
+# --- High-level provisioning -------------------------------------------------
+
+ntfy_provision_publisher() {
+  ntfy_ensure_user "$ntfy_pub_user" || {
+    ntfy_warn "could not create the publisher user; provision manually."
+    return 1
+  }
+  ntfy_grant_acl "$ntfy_pub_user" write-only
+  if ntfy_notify_env_has_token; then
```
kryota-dev: [SHOULD] Publisher half does not self-heal after a `user.db` loss. `ntfy_notify_env_has_token` only checks the token's *format*, not that the server still knows it. When the ntfy state DB is rebuilt, `ntfy_ensure_user` recreates `publisher`, but the stale token in notify-env still passes this gate, so no replacement is minted — the hook then publishes with a dead token while provisioning reports success. The subscriber half already self-heals (re-applies the stored password); mirror that here: reissue the token when the publisher user had to be (re)created (or validate the token against the server before trusting it).

home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+ntfy_ensure_item() {
+  ntfy_item_exists && return 0
+  ntfy_log "creating 1Password item '${ntfy_op_item}'..."
+  op item create --category "Secure Note" --title "$ntfy_op_item" --vault "$ntfy_op_vault" \
+    "subscriber-username[text]=${ntfy_sub_user}" \
+    "subscriber-password[password]=${ntfy_placeholder}" >/dev/null
+}
+
+ntfy_read_subscriber_password() {
+  op read "op://${ntfy_op_vault}/${ntfy_op_item}/subscriber-password" 2>/dev/null || true
```
kryota-dev: [SHOULD] A transient `op read` failure is swallowed to empty and misread as "unprovisioned". `2>/dev/null || true` returns `""` on any error, and `ntfy_provision_subscriber` treats an empty value as "not yet provisioned" → it generates a new password, `change-pass`es the user, and overwrites 1Password — silently locking out every logged-in device. This runs on the apply path every time the lib hash changes, so a biometric re-auth / session timeout between `op item get` (succeeds) and `op read` (fails) is enough to trigger it. Capture `op read`'s exit status and skip the regenerate/overwrite on error (only an absent field / placeholder should trigger generation).

home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+ntfy_magicdns() {
+  local b json dns
+  b="$(ntfy_tailscale_bin)"
+  [ -n "$b" ] || return 1
+  json="$("$b" status --json 2>/dev/null || true)"
+  [ -n "$json" ] || return 1
+  if command -v jq >/dev/null 2>&1; then
+    dns="$(printf '%s' "$json" | jq -r '.Self.DNSName // empty' 2>/dev/null || true)"
+  else
+    dns="$(printf '%s' "$json" | grep -oE '"DNSName"[[:space:]]*:[[:space:]]*"[^"]*"' | head -n 1 | sed 's/.*"\(.*\)"/\1/')"
```
kryota-dev: [NITS] jq-less fallback can pick a peer's DNSName. `grep -oE '"DNSName"…' | head -n1` takes the first `DNSName` in `tailscale status --json`, but key order (`.Self` before `.Peer[]`) is not guaranteed. If a peer's name comes first, `ntfy_print_device_info` / the smoke test would show a wrong server URL. Display is a human-facing hint only (no auto-send), so impact is low — but scoping the fallback to `.Self` (or warning instead of guessing when jq is absent) removes the footgun.

home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+    fi
+    sleep 1
+  done
+  return 1
+}
+
+# --- User / ACL / token primitives ------------------------------------------
+
+ntfy_user_exists() {
+  ntfy_exec user list 2>/dev/null | grep -q "user ${1}"
```
kryota-dev: [NITS] `ntfy_user_exists` uses a substring match: `grep -q "user ${1}"` would also match `user publisher-foo` for `publisher`. Usernames are fixed constants (`publisher`/`subscriber`) and not external input, so there is no exploit — purely robustness. `grep -qw` or an anchored pattern makes it exact.

tests/files.bats
```diff
@@ -1910,32 +1910,80 @@ _gate_decision() {
...
+  grep -qF '.config/ntfy/lib.sh' "$n"
+  # full flow: brings the server up (not check-only) so it can restart a downed container
+  grep -qF 'ntfy_start_server' "$n"
+  grep -qF 'ntfy_assert_serve' "$n"
+  # rotation entry point for both credentials
+  grep -qF -- '--rotate' "$n"
+  grep -qF 'ntfy_rotate_publisher' "$n"
+  grep -qF 'ntfy_rotate_subscriber' "$n"
+  # fail-clear: a distinct exit code when it cannot fix docker/tailscale itself
+  grep -qF 'EXIT_PREREQ' "$n"
```
kryota-dev: [SHOULD] The R-008 interactive rotate guard is entirely unpinned. This block asserts `ntfy_rotate_*` exist, but nothing greps `ntfy_creds_exist`, `ntfy_prompt_yn`, or `/dev/tty` — so the PR's central safety design ("when creds already exist and no `--rotate` is given, ask on /dev/tty; default N without a tty") could regress undetected. Add structural greps for those symbols; ideally `source` the lib and assert `ntfy_prompt_yn` returns non-zero with no tty attached.

tests/files.bats
```diff
@@ -1910,32 +1910,80 @@ _gate_decision() {
...
-  ! grep -qF 'credential[concealed]' "$s"
+  # fail-open: the apply path only ever exits 0 (never hard-fails)
+  ! grep -qE '(^|[[:space:];])exit 1([[:space:];]|$)' "$s"
+}
+
+@test "ntfy shared library provisions and rotates with secret hygiene" {
+  local l="${HOME_DIR}/dot_config/ntfy/lib.sh.tmpl"
+  [ -f "$l" ]
+  # template-stripped body must still parse
+  sed '/{{/d' "$l" | bash -n
```
kryota-dev, line 1948: [SHOULD] This cannot catch the #357 parse-error class it is meant to guard. `sed '/{{/d' | bash -n` strips every line containing a template action *before* the syntax check — which is exactly where a stray literal `{{` in a comment (the #357 regression) would live, so that failure is structurally invisible here and only the CI `chezmoi apply` would catch it. requirements.md mandates "the #357 parse-error regression must not recur". Add a real render assertion — `chezmoi execute-template --source "${HOME_DIR}" <"$l" >/dev/null` (precedent: the claude-plugin-reconciler test) — for lib.sh.tmpl and run_onchange_after_31.

tests/files.bats
```diff
@@ -1983,16 +2031,14 @@ _gate_decision() {
...
-  grep -qF 'onepasswordRead "op://kryota.dev/Dotfiles - ntfy/base-url"' \
-    "${HOME_DIR}/dot_config/ntfy/private_server.yml.tmpl"
+@test "1password ITEMS no longer references the ntfy item" {
+  # base-url was dropped; the ntfy item (read-only subscriber username/password)
+  # is provisioned by ntfy-setup after Tailscale/Docker come up, a later phase
+  # than the validate gate, so the gate must validate nothing ntfy — otherwise a
+  # fresh apply would hard-fail before the item exists.
+  ! _onepassword_item_list | grep -qF 'op://kryota.dev/Dotfiles - ntfy/'
+  # server.yml no longer reads any ntfy field from 1Password.
+  ! grep -qF 'onepasswordRead' "${HOME_DIR}/dot_config/ntfy/private_server.yml.tmpl"
```
kryota-dev, line 2059: [SHOULD] The base-url format-validation stanza removal (R-003) has no regression guard. This test checks the ITEMS array no longer references ntfy, but nothing asserts the `https://*.ts.net` format-check stanza is gone from run_once_after_11-validate-1password. If that stanza is reintroduced, `op read` returns empty, the `case` fails to match, and every fresh `chezmoi apply` hard-fails before the item exists. Add a `! grep` for the removed stanza (e.g. the `ts.net` check / `ntfy_url=`) in the validate script.

tests/files.bats
```diff
@@ -1910,32 +1910,80 @@ _gate_decision() {
...
+  grep -qF 'token remove' "$l"
+  # serve --bg lives here now; funnel is forbidden across the ntfy assets
+  grep -qF 'serve --bg' "$l"
+  ! grep -qE '^[^#]*tailscale.*funnel' "$l"
+  # the removed token-in-1Password / subscriber-token paths must not resurrect
+  ! grep -qF 'credential[concealed]' "$l"
+  ! grep -qF 'subscriber-token' "$l"
+}
+
+@test "ntfy-setup command is on PATH, sources the lib, and rotates" {
```
kryota-dev: [NITS] Test name overclaims: it says "is on PATH" but the body never asserts `~/.local/bin` is on PATH (no such assertion exists anywhere in the suite). Either rename to match what it checks ("exists, sources the lib, and rotates") or add a real assertion that the deploy dir is wired into PATH via the zsh config.

tests/files.bats
```diff
@@ -1910,32 +1910,80 @@ _gate_decision() {
...
   grep -qF 'upstream-base-url: "https://ntfy.sh"' "$y"
-  # tailscale funnel must never appear anywhere in the ntfy config assets.
-  ! grep -ri 'funnel' "${HOME_DIR}/dot_config/ntfy"
+  # base-url was dropped (optional for core messaging); the server config no
+  # longer reads anything from 1Password.
+  ! grep -qE '^[[:space:]]*base-url:' "$y"
+  ! grep -q 'onepasswordRead' "$y"
+  # tailscale funnel must never be USED in the ntfy config assets. A comment that
+  # warns against it (as the lib does) is fine — only non-comment lines count.
+  [ -z "$(grep -rn 'funnel' "${HOME_DIR}/dot_config/ntfy" | grep -v '#')" ]
```
kryota-dev: [NITS] Inconsistent funnel check. This line uses `grep -v '#'` (drops any line that contains a `#` anywhere, so `tailscale funnel status  # debug` would slip through), while the lib test at line 1965 uses the stricter anchored `^[^#]*tailscale.*funnel` (only the pre-comment part). Unify on the anchored form so both defenses are equally robust.

install/install.sh
```diff
@@ -55,3 +55,14 @@ sh -c "$installer" -- init --apply kryota-dev
 echo ""
 echo "=== Setup complete! ==="
 echo "Please restart your terminal."
+
+# ntfy notifications are macOS-only and inherently two-phase: at first apply
+# Tailscale is not yet `up` and Docker Desktop is not running, so the server and
+# device credentials cannot be provisioned during bootstrap. Point the user at
+# the one command that finishes the job once those are available.
+if [ "$OS" = "Darwin" ]; then
```
kryota-dev, line 63: [SHOULD] The Darwin two-phase note (R-009) is untested. The only bootstrap test asserts the file exists; this new `if [ "$OS" = "Darwin" ]` block that points the user at `ntfy-setup` after Tailscale/Docker come up could be removed or garbled without any test catching it. Pin it: assert the ntfy-setup guidance is printed on Darwin and that bootstrap does no provisioning (`! grep -qF 'docker compose' install/install.sh`).

home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+}
+
+# Prompt on /dev/tty; default to N (skip) when no tty is available (fail-open).
+ntfy_prompt_yn() {
+  local ans=""
+  if [ -r /dev/tty ] && [ -w /dev/tty ]; then
+    printf '%s' "$1" >/dev/tty
+    read -r ans </dev/tty || ans=""
+  fi
+  case "$ans" in
```
kryota-dev, line 377: [SHOULD] shfmt style violation that CI silently missed. `shfmt -d -i2 -ci` (the exact flags `make lint`/CI use) wants the `case` labels indented one level under `case`/`esac`; this block (and the `case` blocks in `executable_ntfy-setup`) aren't. It slipped through two lint-gate gaps I verified: (1) the Makefile lint loop `find … | while read -r f; do sed '/{{/d' "$f" | shfmt -d; done` doesn't propagate the per-file non-zero exit, so it prints the diff yet still reaches "All lint checks passed"; (2) the glob `\( -name '*.sh' -o -name '*.sh.tmpl' \)` never matches the extension-less `executable_ntfy-setup`, so that file is neither shellchecked nor shfmt-checked at all. Reformat both files and close the gap (propagate the loop failure + extend the glob to `dot_local/**/executable_*`).

home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+
+ntfy_read_subscriber_password() {
+  op read "op://${ntfy_op_vault}/${ntfy_op_item}/subscriber-password" 2>/dev/null || true
+}
+
+ntfy_store_subscriber_password() {
+  op item edit "$ntfy_op_item" --vault "$ntfy_op_vault" \
+    "subscriber-username[text]=${ntfy_sub_user}" \
+    "subscriber-password[password]=${1}" >/dev/null
+}
```
kryota-dev, line 234: [NITS] The rationale here (and in the PRD, AC-006) — "op has no stdin path for single-field edits" — overstates the constraint. `op item edit --help` documents a stdin/JSON-template path (`op item edit oldLogin --template=file` and `cat updated.json | op item edit oldLogin`). The accepted argv trade-off is still a reasonable pragmatic call for a local one-shot edit, but the precise framing is "no single-field stdin shortcut (would require constructing a full JSON template)", not "no stdin path".

home/dot_config/ntfy/lib.sh.tmpl
```diff
@@ -0,0 +1,365 @@
...
+
+ntfy_read_notify_env_token() {
+  [ -f "$ntfy_env_file" ] || return 1
+  sed -n "s/^NTFY_TOKEN='\{0,1\}\(tk_[A-Za-z0-9]*\).*/\1/p" "$ntfy_env_file" | head -n 1
+}
+
+ntfy_notify_env_has_token() {
+  [ -f "$ntfy_env_file" ] &&
+    grep -qE '^NTFY_TOKEN=.*tk_' "$ntfy_env_file" &&
+    ! grep -qE '^NTFY_TOKEN=.*tk_REPLACE' "$ntfy_env_file"
```
kryota-dev: [NITS] Dead branch: nothing in the current code ever writes a `tk_REPLACE`-shaped token into notify-env (`ntfy_write_notify_env` only writes real server-issued `tk_…` tokens; grep confirms no other writer). This `! grep tk_REPLACE` guard is vestigial from the pre-SSOT script. Since this PR extracted/rewrote this function into the new lib, drop the unreachable check or add a one-line note that it guards a manual-bootstrap convention (if one is intended).

home/dot_local/bin/executable_ntfy-setup
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
```
kryota-dev: [NITS] Shebang drift: this on-PATH executable uses `#!/bin/bash` while the repo's other on-PATH executables (`dot_local/launchers/executable_claude`, `executable_codex`) use `#!/usr/bin/env bash`. Harmless here (no bash-4+ constructs), but `#!/usr/bin/env bash` matches the established convention for this directory family.

kryota-dev (commented):
## Multi-Review 補足（5 本目 cc-code-review の net-new 指摘）

先の統合レビュー（pullrequestreview-4781559981）は 5 本目の cc-code-review 完了前に投稿してしまったため、その net-new 指摘を追補します。すべて一次検証済み（shfmt / make lint / op item edit --help / grep）。

### severity 引き上げ

- **user.db 消失後の publisher 非再発行を MUST へ引き上げ**（既投稿 lib.sh.tmpl:232 と同一）。cc-code-review が doc 矛盾を追加検証: `docs/architecture/notifications.md:186-187` は「delete user.db → run ntfy-setup; it recreates both users and re-applies the known password/token」と明記するが、これは **publisher に対して現状偽**（subscriber は reconcile branch で真）。コード修正（publisher user を再作成した場合は token 再発行）で doc の主張を真にして解消する。

### net-new [SHOULD]

- **新規ファイルの shfmt 違反が lint ゲートの穴で見逃されている**（インライン: lib.sh.tmpl の case ブロック）。`shfmt -d -i2 -ci` で lib.sh.tmpl（`ntfy_prompt_yn` の case ラベル未インデント）と executable_ntfy-setup（case ラベル + コメント整列）に diff。見逃し経路 2 つを実体確認: (1) `Makefile` の lint ループ `find … | while read -r f; do sed '/{{/d' "$f" | shfmt -d; done` が per-file exit status を伝播せず、diff を出しても "All lint checks passed" に到達。(2) glob `\( -name '*.sh' -o -name '*.sh.tmpl' \)` が拡張子なしの `executable_ntfy-setup` を一切マッチせず shellcheck/shfmt 対象外（launchers/executable_{claude,codex} と同じ既存の穴に 3 本目を追加）。→ 両ファイルを整形 + Makefile ループの失敗伝播 + glob 拡張で穴を塞ぐ。
- **server.yml の陳腐化コメントが本 PR で編集中のファイルに残存**（`private_server.yml.tmpl:32-34`、hunk 外のため body で報告）。`# Users/tokens are provisioned once manually via docker compose exec` は #357 の自動プロビジョニング導入時点で既に虚偽。AC-011 の「stale な manual provisioning framing を一掃」の意図に合致するため本 PR で修正。
- **ci-and-tests.md（EN/JA）の AC-011 未反映を再確認**（architecture-reviewer と収束）。`git log` で本ファイルが #357 以来未編集・本 PR 差分に不在を確認。除外リストを「1Password-backed secret template」で括る framing が base-url 除去後の server.yml（もう op-backed でない）に対して不正確。

### net-new [NITS]

- **op stdin 主張の精度**（インライン: lib.sh.tmpl:222 コメント / PRD AC-006）。`op item edit --help`（実体確認）は `--template=file` / `cat json | op item edit` の stdin/JSON テンプレート経路を持つ。「op offers no stdin path」は言い過ぎで、正確には「単一フィールドの stdin ショートカットが無い」。argv trade-off を受容する判断自体は妥当だが framing を是正。
- **tk_REPLACE 死にロジック**（インライン: lib.sh.tmpl:195）。`ntfy_notify_env_has_token` の `! grep tk_REPLACE` は notify-env に `tk_REPLACE` を書く経路が存在しない（grep 確認済み）ため到達不能な防御。pre-PR からの vestigial。
- **shebang 不一致**（インライン: executable_ntfy-setup:1）。`#!/bin/bash` は launchers/executable_{claude,codex} の `#!/usr/bin/env bash` と不揃い（実害なし・bash4 構文不使用）。

### GOOD（cc-code-review が実挙動で追検証）

- 1Password の `[password]` field-type が `op item create/edit --dry-run --format json`（live vault）で `"type": "CONCEALED"` に解決することを確認 — docs の「(concealed)」主張どおりでセキュリティ退行なし。
- rotation 順序（read old → issue new → write → remove old）、secret hygiene（env/heredoc のみ・set -x 皆無）、SSOT 抽出、bilingual docs、全テスト緑（232/232）を実行確認。

_これら net-new は review-resolve-loop で既投稿 11 件と合わせて一括対応する。_

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — `ntfy_rotate_publisher` now returns non-zero when `ntfy token remove` fails (so a possibly-still-valid old token is surfaced) instead of logging success, and the lib header now names `ntfy token remove` as the second accepted local/transient argv exposure.

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — `ntfy_provision_publisher` now records whether the publisher user pre-existed and reissues the token when it had to be (re)created, so "delete user.db, run ntfy-setup" restores the publisher too (mirrors the subscriber self-heal). This also makes the notifications.md recovery claim accurate.

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — `ntfy_read_subscriber_password` no longer swallows the op exit status; `ntfy_provision_subscriber` now bails out and leaves the credential untouched on a real op error, so a transient failure can no longer overwrite a valid subscriber password.

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — the jq-less MagicDNS fallback now strips up to the `"Self"` object before taking the first DNSName, so a peer's name can no longer be picked.

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — `ntfy_user_exists` now matches the name exactly (`grep -qE "user NAME([ :(]|$)"`), so `publisher` cannot match a hypothetical `publisher-foo`.

tests/files.bats
kryota-dev: Fixed in 525039e — the lib and ntfy-setup tests now pin the R-008 interactive guard (`ntfy_creds_exist`, `ntfy_prompt_yn`, `/dev/tty`).

tests/files.bats
kryota-dev: Fixed in 525039e — added a real `chezmoi execute-template --source` render assertion for lib.sh.tmpl, run_onchange_after_31, and server.yml, which catches the #357 parse-error class the stripped `bash -n` cannot.

tests/files.bats
kryota-dev: Fixed in 525039e — the ITEMS test now also asserts the base-url format-validation stanza (`ts.net` / `ntfy_url=`) is absent from run_once_after_11-validate-1password.

tests/files.bats
kryota-dev: Fixed in 525039e — renamed the test to describe what it actually checks and added a `.chezmoiignore` deploy-path assertion (the wiring that puts it on PATH).

tests/files.bats
kryota-dev: Fixed in 525039e — the funnel check now strips inline comments (`sed 's/#.*//'`) before matching, so a trailing `# … funnel …` note can't mask a real usage — consistent with the lib test's anchored form.

install/install.sh
kryota-dev: Fixed in 525039e — added a test asserting the Darwin-only ntfy-setup note and that bootstrap performs no provisioning (`! grep docker compose / tailscale serve / op item`).

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — reformatted lib.sh.tmpl and executable_ntfy-setup with `shfmt -i2 -ci`, made the Makefile lint loops propagate per-file failures (`|| exit 1`), and extended the glob to cover `home/dot_local/bin/executable_*` (the command was previously unlinted). `make lint` stays green.

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — reworded to "no single-field stdin shortcut" (a full JSON-template stdin path does exist) in the lib header and the secrets-and-isolation docs.

home/dot_config/ntfy/lib.sh.tmpl
kryota-dev: Fixed in 525039e — dropped the unreachable `tk_REPLACE` check; only real server-issued `tk_…` tokens are ever written to notify-env, so a shape check suffices.

home/dot_local/bin/executable_ntfy-setup
kryota-dev: Fixed in 525039e — changed the shebang to `#!/usr/bin/env bash` to match the other on-PATH executables.

PR #358, opened 2026-07-26
feat(claude,codex): unify the launcher families into PATH wrappers with built-in env injection

kryota-dev:
## Why

`cld` / `cld-r06` / `cdx` / `cdx-r06` were interactive-zsh **aliases**, so non-interactive contexts
(hooks, launchd, scripts, Claude's Bash tool) never saw them and resolved the bare `claude`/`codex`
binary with no per-account isolation env. That forced env injection to be hand-copied in several
places — morning-radar's "keep in sync with `_claude_with_home`" block (where #341 found a real bug)
and the codex skill's inline account-selection prelude, repeated 11×. `secrets-and-isolation.md`
named this "two places that must stay in sync" as the isolation model's main maintenance burden.

This unifies env injection into wrapper commands on PATH so **every** launch — interactive,
hook-spawned, launchd, or Bash-tool — goes through one place. It takes the low-risk path
(intent-gated with the maintainer): claude stays mise-managed, and the `~/.local/bin/claude`
native-install self-check symlink, the `mise exec -- claude` provisioning convention, and
`run_once_after_16` are all untouched — so there is no fresh-machine-bootstrap or bricked-launch
risk. (Approach "disable claude in mise + wrapper replaces the symlink" was considered and rejected:
adversarial review surfaced an unverified cold-install path, pin-incorrect binary resolution, and an
unverifiable native-self-check assumption.)

## What changed

**New launcher wrappers** (`home/dot_local/launchers/`)
- `executable_claude` / `executable_codex` — real scripts that dispatch on `$0` and inject the
  per-account env, then exec the real binary (claude via `mise which claude`, pin-correct; codex via
  `/opt/homebrew/bin/codex`). `symlink_{cld,cld-r06,cdx,cdx-r06}` point at them.
- **Account rule:** bare `claude`/`codex` fill-gaps — a hook-spawned child keeps its parent
  session's account; `cld-r06`/`cdx-r06` override unconditionally. For codex, `CLAUDE_CONFIG_DIR` is
  authoritative and a mismatched inherited `CODEX_HOME` is overridden, closing a cross-account leak.
- **codex `--profile shared`** is injected only when argv carries no `--profile` (per-token scan
  recognising `--profile`/`--profile=…`/`-p`/`-p…`, stopping at `--`, placed at the global position
  so `exec` / `exec review` forms parse); `--profile agent` passes through unchanged.
- The wrapper sources the 0600 MCP-keys file itself, scoped to its own process, and fails loudly if
[...]

PR #373, opened 2026-07-26
feat(brew): auto-dump the Brewfile via a PATH-shim wrapper

kryota-dev:
## Summary

Replace the manual `make dump-brewfile` target with a `brew` **PATH-shim wrapper** that keeps the
tracked Brewfile in sync automatically. After a package-set-mutating subcommand
(`install`/`uninstall`/`reinstall`/`tap`/`untap`), the wrapper regenerates the chezmoi source
`dot_Brewfile` via `brew bundle dump`, so it no longer drifts after ad-hoc `brew install` /
`brew uninstall`.

Homebrew has no post-install/post-uninstall hook (Homebrew/brew#2202, open since 2017), and a
shell-function wrapper would only cover interactive shells. A real file ahead of the real `brew` on
`PATH` (`~/.local/launchers`, already prepended in `.zprofile`/`.zshrc`) fires from any shell that
resolves through that dir, following the launcher PATH-wrapper pattern from #358.

Closes #360.

## Changes

- **New:** `home/dot_local/launchers/executable_brew` — runs the real brew off-`PATH`, preserves its
  exit code, and dumps the Brewfile after a mutating subcommand.
- **New:** `tests/brew_launcher.bats` — stub-based regression tests (no real brew/chezmoi needed).
- **Removed:** the `make dump-brewfile` target (`.PHONY` + recipe).
- **Docs:** dropped every `dump-brewfile` reference (`README.md`/`.ja.md`, `AGENTS.md`,
  `docs/contributing/local-dev.md`/`.ja.md`, `docs/explanation/design-rationale.md`/`.ja.md`) and
  rewrote the Brewfile paragraph in `docs/architecture/dev-tooling.md`/`.ja.md` to document the
  wrapper as the new sync mechanism.

## Design decisions

- **Brewfile resolution:** `"$(chezmoi source-path)/dot_Brewfile"` — cwd-independent; the dump is
  skipped (with a stderr note) when chezmoi is not installed.
- **Synchronous dump:** runs after the real brew exits and never masks the real brew's exit code
  (a dump failure only warns).
- **Real-brew resolution off-`PATH`:** `BREW_LAUNCHER_BIN` (test seam) → `$HOMEBREW_PREFIX` →
  `/opt/homebrew` → `/usr/local` → `/home/linuxbrew/.linuxbrew`; fails loudly (exit 127) if none
  resolve. Absolute paths never re-enter the wrapper, so there is no recursion.

## Known limitation (accepted)

Absolute-path invocations (`/opt/homebrew/bin/brew`) bypass the shim. `run_onchange_before_10-brew-bundle`
runs `brew bundle` (not a mutating subcommand), so it never triggers a dump — no recursion.

## Testing

- `make lint` — pass (shellcheck + shfmt + zsh syntax).
- `bats tests/*.bats` — 237 pass / 0 fail (includes the new `brew_launcher.bats` and `docs_facts.bats`,
  which stays green after the target removal).
- New tests cover: the trigger set, a global flag before the subcommand, non-mutating subcommands not
  dumping, exit-code fidelity (42 → 42), the chezmoi-absent skip, and fail-loud real-brew resolution.

### DIFF
--- a/home/dot_local/launchers/executable_brew
+++ b/home/dot_local/launchers/executable_brew
@@ -18,15 +18,21 @@
 # (not a mutating subcommand), so it never triggers a dump — no recursion.
 set -eu
 
-# Subcommands that change the installed package set and so warrant a Brewfile refresh.
+# Subcommands that change the installed package set and so warrant a Brewfile refresh. `remove` and
+# `rm` are aliases of `uninstall` (see `brew help uninstall`), so they must trigger too — otherwise
+# `brew rm foo` would drift the Brewfile, the very problem this wrapper fixes. install/reinstall/
+# tap/untap have no aliases.
 is_dump_trigger() {
   case "$1" in
-    install | uninstall | reinstall | tap | untap) return 0 ;;
+    install | uninstall | remove | rm | reinstall | tap | untap) return 0 ;;
     *) return 1 ;;
   esac
 }
 
 # The subcommand is the first non-option argument (global flags like `--verbose` may precede it).
+# This is a heuristic, not brew's full dispatcher: a query form whose first non-option token happens
+# to be a trigger word (e.g. `brew --prefix install`, where `install` is a formula name) runs one
+# extra, harmless dump — the dump faithfully reflects the unchanged package set. Accepted limitation.
 subcommand=""
 for arg in "$@"; do
   case "$arg" in
@@ -39,13 +45,18 @@ for arg in "$@"; do
 done
 
 # Resolve the real brew WITHOUT going through PATH (which would re-resolve this wrapper).
-# BREW_LAUNCHER_BIN overrides the resolver for tests. Otherwise prefer the exported HOMEBREW_PREFIX,
-# then the standard install prefixes for Apple Silicon, Intel, and Linuxbrew. Fail loudly rather
-# than silently do nothing.
+# BREW_LAUNCHER_BIN overrides the resolver for tests. HOMEBREW_PREFIX is exported into every shell
+# by `brew shellenv`, so later code could repoint it at an attacker-controlled dir; trust it only
+# when it names a standard prefix (allowlist), then fall back to the standard install prefixes for
+# Apple Silicon, Intel, and Linuxbrew. Fail loudly rather than silently do nothing.
 real="${BREW_LAUNCHER_BIN:-}"
 if [ -z "$real" ]; then
+  case "${HOMEBREW_PREFIX:-}" in
+    /opt/homebrew | /usr/local | /home/linuxbrew/.linuxbrew) prefix_cand="$HOMEBREW_PREFIX/bin/brew" ;;
+    *) prefix_cand="" ;;
+  esac
   for cand in \
-    "${HOMEBREW_PREFIX:+$HOMEBREW_PREFIX/bin/brew}" \
+    "$prefix_cand" \
     /opt/homebrew/bin/brew \
     /usr/local/bin/brew \
     /home/linuxbrew/.linuxbrew/bin/brew; do
@@ -66,16 +77,27 @@ status=0
 "$real" "$@" || status=$?
 
 # After a package-set-mutating subcommand, refresh the chezmoi source Brewfile so it tracks the
-# ad-hoc change. Resolve the target via `chezmoi source-path` (cwd-independent). A dump failure or a
-# missing chezmoi only warns — the caller still sees the real brew's original exit code.
+# ad-hoc change. Resolve the target via `chezmoi source-path` (cwd-independent). Every failure path
+# only warns and still falls through to `exit "$status"`, so the caller always sees the real brew's
+# original exit code (AC3). The guarded `if source_path="$(...)"` matters: an unguarded assignment
+# would abort here under `set -e` on a chezmoi failure and mask that exit code.
 if is_dump_trigger "$subcommand"; then
-  if command -v chezmoi >/dev/null 2>&1; then
-    brewfile="$(chezmoi source-path)/dot_Brewfile"
-    if ! "$real" bundle dump --force --file="$brewfile" >/dev/null 2>&1; then
-      echo "brew launcher: 'brew bundle dump' failed; $brewfile not refreshed." >&2
-    fi
-  else
+  if ! command -v chezmoi >/dev/null 2>&1; then
     echo "brew launcher: chezmoi not found; skipped Brewfile refresh." >&2
+  elif ! source_path="$(chezmoi source-path)"; then
+    echo "brew launcher: 'chezmoi source-path' failed; Brewfile not refreshed." >&2
+  else
+    brewfile="$source_path/dot_Brewfile"
+    # Neutralize the subtractive HOMEBREW_BUNDLE_DUMP_NO_* vars so a stray env setting cannot make
+    # the automatic --force dump overwrite the tracked Brewfile with a whole category missing.
+    # Capture the dump's own stderr so a failure reports its cause, not just a generic line.
+    if ! dump_err="$(env \
+      -u HOMEBREW_BUNDLE_DUMP_NO_BREW -u HOMEBREW_BUNDLE_DUMP_NO_CASK -u HOMEBREW_BUNDLE_DUMP_NO_TAP \
+      -u HOMEBREW_BUNDLE_DUMP_NO_MAS -u HOMEBREW_BUNDLE_DUMP_NO_VSCODE \
+      "$real" bundle dump --force --file="$brewfile" 2>&1 >/dev/null)"; then
+      echo "brew launcher: 'brew bundle dump' failed; $brewfile not refreshed." >&2
+      if [ -n "$dump_err" ]; then printf '%s\n' "$dump_err" >&2; fi
+    fi
   fi
 fi
 
--- a/home/dot_local/launchers/executable_brew
+++ b/home/dot_local/launchers/executable_brew
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# PATH-shim wrapper for `brew` that keeps the chezmoi-managed Brewfile in sync. After a
+# package-set-mutating subcommand (install/uninstall/reinstall/tap/untap) it regenerates the
+# chezmoi source dot_Brewfile via `brew bundle dump`, so the tracked Brewfile no longer drifts
+# after ad-hoc `brew install` / `brew uninstall` the way it did when only `make dump-brewfile`
+# refreshed it.
+#
+# Homebrew has no post-install/post-uninstall hook (Homebrew/brew#2202, open since 2017). A shell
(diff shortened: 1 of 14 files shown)


home/dot_local/launchers/executable_brew
```diff
@@ -0,0 +1,82 @@
...
+# Run the real brew, preserving its exit code for the caller (works under set -e).
+status=0
+"$real" "$@" || status=$?
+
+# After a package-set-mutating subcommand, refresh the chezmoi source Brewfile so it tracks the
+# ad-hoc change. Resolve the target via `chezmoi source-path` (cwd-independent). A dump failure or a
+# missing chezmoi only warns — the caller still sees the real brew's original exit code.
+if is_dump_trigger "$subcommand"; then
+  if command -v chezmoi >/dev/null 2>&1; then
+    brewfile="$(chezmoi source-path)/dot_Brewfile"
```
kryota-dev: [MUST] Under `set -eu`, this unguarded command-substitution assignment masks the real brew's exit code: if `chezmoi source-path` exits non-zero, errexit aborts before `exit "$status"` (AC3 violation). Verified with a bash repro. Guard it: `if source_path="$(chezmoi source-path)"; then brewfile="$source_path/dot_Brewfile"; ...; else echo 'brew launcher: chezmoi source-path failed; Brewfile not refreshed.' >&2; fi`.

home/dot_local/launchers/executable_brew
```diff
@@ -0,0 +1,82 @@
...
+#
+# Known limitation (accepted): an absolute-path invocation (/opt/homebrew/bin/brew) bypasses the
+# shim, so its mutations are not auto-dumped. run_onchange_before_10-brew-bundle runs `brew bundle`
+# (not a mutating subcommand), so it never triggers a dump — no recursion.
+set -eu
+
+# Subcommands that change the installed package set and so warrant a Brewfile refresh.
+is_dump_trigger() {
+  case "$1" in
+    install | uninstall | reinstall | tap | untap) return 0 ;;
```
kryota-dev: [SHOULD] `brew uninstall` has the aliases `remove` and `rm` (verified via `brew help uninstall`). `brew rm foo` / `brew remove foo` uninstall but don't trigger a dump, so the Brewfile still drifts — the exact problem this PR fixes. Add `remove | rm` to the trigger set. (`install`/`reinstall`/`tap`/`untap` have no aliases.)

home/dot_local/launchers/executable_brew
```diff
@@ -0,0 +1,82 @@
...
+done
+
+# Resolve the real brew WITHOUT going through PATH (which would re-resolve this wrapper).
+# BREW_LAUNCHER_BIN overrides the resolver for tests. Otherwise prefer the exported HOMEBREW_PREFIX,
+# then the standard install prefixes for Apple Silicon, Intel, and Linuxbrew. Fail loudly rather
+# than silently do nothing.
+real="${BREW_LAUNCHER_BIN:-}"
+if [ -z "$real" ]; then
+  for cand in \
+    "${HOMEBREW_PREFIX:+$HOMEBREW_PREFIX/bin/brew}" \
```
kryota-dev: [SHOULD] `HOMEBREW_PREFIX` is used unvalidated to locate the real brew. Marginal risk over existing PATH-manipulation is low (an attacker who can set this env var can also prepend PATH), but an allowlist against `/opt/homebrew` / `/usr/local` / `/home/linuxbrew/.linuxbrew` is cheap and matches the wrapper's off-PATH intent — this repo only uses standard prefixes.

home/dot_local/launchers/executable_brew
```diff
@@ -0,0 +1,82 @@
...
+status=0
+"$real" "$@" || status=$?
+
+# After a package-set-mutating subcommand, refresh the chezmoi source Brewfile so it tracks the
+# ad-hoc change. Resolve the target via `chezmoi source-path` (cwd-independent). A dump failure or a
+# missing chezmoi only warns — the caller still sees the real brew's original exit code.
+if is_dump_trigger "$subcommand"; then
+  if command -v chezmoi >/dev/null 2>&1; then
+    brewfile="$(chezmoi source-path)/dot_Brewfile"
+    if ! "$real" bundle dump --force --file="$brewfile" >/dev/null 2>&1; then
```
kryota-dev: [SHOULD] `$HOMEBREW_BUNDLE_DUMP_NO_BREW`/`_NO_CASK`/`_NO_TAP` exist (verified in `brew bundle dump --help`). If set, this automatic `--force` dump overwrites the tracked Brewfile omitting that category — silent, recurring data loss amplified by the automation. Neutralize them for the dump: `env -u HOMEBREW_BUNDLE_DUMP_NO_BREW -u HOMEBREW_BUNDLE_DUMP_NO_CASK -u HOMEBREW_BUNDLE_DUMP_NO_TAP -u HOMEBREW_BUNDLE_DUMP_NO_MAS -u HOMEBREW_BUNDLE_DUMP_NO_VSCODE "$real" bundle dump ...`.

[NITS] `>/dev/null 2>&1` also hides the dump's own error on failure — capture and forward stderr so the generic warning has a cause.

kryota-dev (commented):
## Multi-Review consolidated result — PR #373

Four reviewers ran in parallel: cc-code-review, cc-security-review, test-reviewer, and Codex (cross-model). CI is green (7 checks). Every load-bearing claim below was fact-checked against real `brew` / `bash` behavior before posting.

### Summary (post fact-check)

| Category | Count |
|---|---|
| MUST | 1 |
| SHOULD | 3 |
| NITS | 3 |
| GOOD | recursion analysis, arg-array quoting (no injection), `--force` atomic overwrite, doc-removal completeness, exit-code capture pattern |

### [MUST] `chezmoi source-path` failure masks the real brew exit code (AC3)
`home/dot_local/launchers/executable_brew:73` — `brewfile="$(chezmoi source-path)/dot_Brewfile"` is an unguarded assignment under `set -eu`. If `chezmoi source-path` exits non-zero (corrupt config, etc.), errexit aborts before `exit "$status"`, so the wrapper returns chezmoi's code instead of the real brew's — violating AC3. Found independently by cc-code-review, cc-security-review, and Codex. Verified: `bash -eu -c 'f(){ return 3; }; x="$(f)/s"; echo hi'` exits 3.
Fix: `if source_path="$(chezmoi source-path)"; then brewfile="$source_path/dot_Brewfile"; ... else warn; fi`.

### [SHOULD] `uninstall` aliases `remove` / `rm` are not dump triggers
`executable_brew:24` — verified via `brew help uninstall`: `brew uninstall, remove, rm` is one command. `brew rm foo` uninstalls but never refreshes the Brewfile, re-introducing the drift this PR targets. (`install`/`reinstall`/`tap`/`untap` have no aliases — verified.) Fix: add `remove | rm`.

### [SHOULD] `HOMEBREW_BUNDLE_DUMP_NO_*` can silently truncate the Brewfile on auto-dump
`executable_brew:74` — `$HOMEBREW_BUNDLE_DUMP_NO_BREW` / `_NO_CASK` / `_NO_TAP` exist (verified in `brew bundle dump --help`). If any is set, the automatic `brew bundle dump --force` overwrites the tracked Brewfile omitting that whole category. The old manual `make dump-brewfile` shared this behavior, but firing it automatically on every mutating subcommand makes silent category loss far more likely. Recoverable via git, hence SHOULD. Codex rated it MUST; downgraded here. Fix: neutralize the subtractive env vars for the dump call (`env -u HOMEBREW_BUNDLE_DUMP_NO_BREW -u ... "$real" bundle dump ...`).

### [SHOULD] `HOMEBREW_PREFIX` is trusted without validation (defense-in-depth)
`executable_brew:48` — cc-security-review rated this High. Threat-model note: an attacker able to set `HOMEBREW_PREFIX` can equally `export PATH=/evil:$PATH` to bypass the wrapper entirely, so the marginal risk over existing PATH-manipulation is low (downgraded to SHOULD). Still, an allowlist matches the wrapper's off-PATH intent and is cheap because this repo only uses standard prefixes. Fix: accept `HOMEBREW_PREFIX` only when it equals `/opt/homebrew`, `/usr/local`, or `/home/linuxbrew/.linuxbrew`, falling through to the standard candidates otherwise.

### [NITS]
- `executable_brew:74`: dump failure hides `brew bundle dump`'s own stderr (`>/dev/null 2>&1`), leaving only a generic message — capture and forward it on failure.
- `executable_brew:29`: the "first non-option arg = subcommand" heuristic can false-positive on query forms like `brew --prefix install`. Impact is only a spurious, faithful (no-op) dump — accepted as a documented limitation rather than reproducing full brew dispatch.
- Neither `make lint` nor the bats suite runs `shellcheck` on the launcher (only `bash -n`); the claude/codex launchers share this gap. A shellcheck assertion in `brew_launcher.bats` would guard regressions.

### Test coverage (test-reviewer)
The suite exercises only `install`; the other trigger subcommands, the dump-failure / source-path-failure branches (AC3), and the `HOMEBREW_PREFIX` fallback resolution are uncovered. Tests will be added alongside the fixes.

### Fact-checked → rejected / accepted
- Codex `[MUST] brew --prefix install` → downgraded to NITS (harmless no-op dump).
- cc-security `[Low]` no owner/integrity check on the resolved brew → out of scope for this threat model (reviewer concurs).
- cc-security `[Info]` on `CLAUDE_LAUNCHER_BIN`/`CODEX_LAUNCHER_BIN` → out of scope (#358 files, not this PR).

home/dot_local/launchers/executable_brew
kryota-dev: **Fixed ( 1834102 )** — guarded the assignment as `if source_path="$(chezmoi source-path)"; then brewfile="$source_path/dot_Brewfile"; ...; else warn; fi`, so a chezmoi failure no longer masks the real brew exit code. Added an AC3 regression test (source-path failure preserves the exit code).

home/dot_local/launchers/executable_brew
kryota-dev: **Fixed ( 1834102 )** — added `remove | rm` to the trigger set, with a test covering all seven trigger subcommands.

home/dot_local/launchers/executable_brew
kryota-dev: **Fixed ( 1834102 )** — HOMEBREW_PREFIX is now allowlisted to the standard prefixes (`/opt/homebrew` / `/usr/local` / `/home/linuxbrew/.linuxbrew`) before use, falling through to the standard candidates otherwise. Added a test asserting a non-standard HOMEBREW_PREFIX is not trusted.

home/dot_local/launchers/executable_brew
kryota-dev: **Fixed ( 1834102 )** — the dump now runs under `env -u HOMEBREW_BUNDLE_DUMP_NO_BREW -u ...`, so a stray env setting cannot truncate a category. The dump's own stderr is now forwarded on failure. Both covered by tests.

PR #380, opened 2026-07-26
refactor(skills): 2026-07-27 stocktake — retire cleanup-plan

kryota-dev:
## Summary

Runs a full `skill-stocktake` over the 46 curated skills in `home/dot_agents/skills/`
(holistic evaluation against overlap / uniqueness / currency / actionability, with
per-skill transcript usage counts fed in as evidence) and retires the one skill the
review — and the owner — agreed no longer earns its upkeep.

Closes #369.

**Usage caveat:** transcript counts come from a ~20-day window (`cleanupPeriodDays: 20`),
so `usage=0` means "no recent use", **not** "never used" and not "low value". Usage was
one input to the holistic verdict, never the verdict itself — most zero-use skills are
load-bearing (invoked indirectly by an orchestrator / agent / launchd wrapper) or encode
irreproducible unique value, and were kept.

## What changed

- **Retired `cleanup-plan`** (`git rm`): a 62-line wrapper over a single `> <plan-file>`
  shell redirect with no non-obvious logic; unreferenced by any skill/agent/test/CLAUDE.md
  and unused in the window. The same need (start a fresh plan in-session) is met by a
  manual redirect or re-entering plan mode.
- **Ripple updates** (keep every guard green, following the 2026-07-06 stocktake pattern):
  - `home/.chezmoiremove`: register `.agents/skills/cleanup-plan` so the already-deployed
    copy is removed on the next `chezmoi apply`.
  - `tests/skill_provenance.bats`: add `cleanup-plan` to the retired-skill guard array.
  - `docs/agents/skills-provenance.md` + `.ja.md`: `FACT:curated-skill-count` 46 → 45,
    "Session and context management" theme 6 → 5 skills, and drop `cleanup-plan` from the
    themed list (EN + JA mirror in lock-step).

## Skill-stocktake verdicts (all 46 curated skills)

Satisfies AC-1 ("every curated skill has a recorded verdict"). Evaluated by three
independent `general-purpose` subagents (≤16 skills each), each doing a mandatory
dependency scan (`git grep`) before any Retire/Merge.

| Verdict | Count | Skills |
|---|---|---|
| **Retire (applied)** | 1 | `cleanup-plan` |
| **Retire candidate — declined by owner (kept)** | 4 | `fix-migration-leftover`, `webp-convert`, `github-projects`, `knowledge-distill` |
| **Improve — deferred to follow-up (kept)** | 2 | `planning`, `sync-daily-planning-calendar` |
| **Keep** | 39 | all others |

### Declined retire candidates (owner review)

- `fix-migration-leftover`, `webp-convert` — flagged as thin/reproducible, but the owner
  chose to keep them.
- `github-projects` — only ProjectV2 GraphQL reference; kept (owner did not confirm
  non-use of GitHub Projects V2).
- `knowledge-distill` — only weekly health auditor of the CLV2/ECC continuous-learning
  loop; **kept because CLV2 is active in this environment**.

### Deferred Improve follow-ups (not in this PR)

- `planning` — broken markdown code-fence nesting (~L210-263), a dated "Claude 4" heading,
  and a ~30-line question template duplicated near-verbatim (inflates the file to 614 lines).
- `sync-daily-planning-calendar` — references stale tool names `gcal_list_events` /
  `gcal_create_event` / `gcal_delete_event`; the real MCP names are
  `mcp__claude_ai_Google_Calendar__{list_events,create_event,delete_event}`.

## Verification

- `make test` → **exit 0**, `1..257`, no failures. Key guards:
  - `docs_facts: FACT:curated-skill-count` matches the dir count (45).
  - `docs_facts: every EN doc has a .ja.md mirror`.
  - `skill provenance: retired unmanaged skills are absent from source` (now includes
    `cleanup-plan`).
- The runtime informational check lists `cleanup-plan` as an unmanaged **runtime** skill;
  that is expected until `chezmoi apply` runs the new `.chezmoiremove` entry, and the
  check is informational (exit 0), not a gate.


tests/skill_provenance.bats
```diff
@@ -186,7 +188,7 @@ _skill_is_external() {
     agentcore vercel-sandbox patch-remote-control find-skills \
     self-evaluate ghostwrite-comment spec serena drawio-skill-cli \
     copilot-agent-task open-pr-files vitest-error-analysis \
-    fetch-pr-comments save-session wtp-workspace; do
+    fetch-pr-comments save-session wtp-workspace cleanup-plan; do
```
kryota-dev, line 191: [SHOULD] This retired-skill guard only checks source absence. Unlike the `agent-browser` guard above (which also asserts `grep -qFx ".agents/skills/${s}" "${HOME_DIR}/.chezmoiremove"` for each retired name), it does not verify that a matching `.chezmoiremove` removal target exists — so a future accidental deletion of a `.chezmoiremove` line would go undetected and a retired skill could silently linger on already-provisioned machines.

Consider adding a symmetric `.chezmoiremove` existence assertion (or extracting a shared helper). **Caveat:** scope it to the deployed subset — `patch-remote-control` / `find-skills` are in this array but have no `.chezmoiremove` entry (never deployed under `.agents/skills/<name>`), so a blanket loop over the whole array would fail.

Pre-existing gap, not introduced by this PR; safe to defer to a follow-up.

docs/agents/skills-provenance.md
```diff
@@ -86,11 +86,11 @@ Structured task decomposition, requirement analysis, and end-to-end delivery flo
 
 `planning`, `sdd`, `grill-me`, `prompt-conform`, `pr-workflow`
 
-### Session and context management (6 skills)
+### Session and context management (5 skills)
```
kryota-dev, line 89: [SHOULD] The per-theme `(N skills)` headings and the backtick skill lists here are not machine-verified against `FACT:curated-skill-count`. `docs_facts.bats` checks only the FACT number against the directory count, so a future cross-theme move could update one heading/list without the other and no test would catch it.

Consider an awk/grep test asserting that each theme's `(N skills)` equals its backtick-list length and that all themes sum to the FACT count. Pre-existing gap, not introduced by this PR; safe to defer to a follow-up.

kryota-dev (commented):
## Multi-Review integrated result — PR #380

Four reviewers ran in parallel (`cc-code-review`, `cc-security-review`, `codex`, `test-reviewer`). There were no pre-existing reviews, so nothing was deduplicated.

### Verdict summary

| Category | cc-code-review | cc-security-review | codex | test-reviewer |
|---|---|---|---|---|
| MUST   | 0 | 0 | 0 | 0 |
| SHOULD | 0 | 0 (+1 Info) | 0 | 2 |
| NITS   | 2 | — | 0 | 2 |
| GOOD   | 2 | 5 | — | 2 |

### Security
- Overall risk level: **Low**
- Vulnerabilities: **0** (2 Info)

### [SHOULD] — both are pre-existing structural gaps, not regressions introduced by this PR

1. **The retired-skill guard does not machine-verify the matching `.chezmoiremove` entry** (converged: `cc-security-review` + `test-reviewer`). The `agent-browser` guard (`tests/skill_provenance.bats:44-51`) asserts `grep -qFx ".agents/skills/${s}" .chezmoiremove` for every retired name; the "retired unmanaged skills are absent from source" guard (L178-197, where `cleanup-plan` was added) only checks source absence. A future accidental deletion of a `.chezmoiremove` line would therefore go undetected, and a retired skill could silently linger on already-provisioned machines. **Caveat:** a naive blanket loop would fail — `patch-remote-control` / `find-skills` are in the array but have no `.chezmoiremove` entry (never deployed under `.agents/skills/<name>`), so any assertion must scope to the deployed subset.

2. **Per-theme counts are not machine-verified against `FACT:curated-skill-count`** (`test-reviewer`). `docs_facts.bats` checks only the FACT number against the directory count; the per-theme `(N skills)` headings and the backtick skill lists in `skills-provenance.md` are not asserted to sum to the FACT count, so a future cross-theme move could update one without the other.

### [NITS]
- The informational runtime check lists `cleanup-plan` among 6 unmanaged runtime skills during `make test` — expected until `chezmoi apply` runs the new `.chezmoiremove` entry (exit 0, non-blocking). Flagged so it is not misread as breakage.
- The retired-guard array is append-ordered (not alphabetized); harder to eyeball for double-registration as it grows.
- Comment claims ("62-line wrapper", "unreferenced / unused") were independently verified as accurate.

### [GOOD]
- The integrity work is thorough and cross-consistent: FACT marker, themed EN/JA lists, `.chezmoiremove`, and the retired-guard array were all cross-checked against real data with zero discrepancy.
- The `.chezmoiremove` comment records *why* `git rm` alone is insufficient (already-deployed copies persist), following the 2026-07-06 stocktake pattern.
- EN/JA docs were updated in lock-step; `FACT:curated-skill-count` (45) matches the actual directory count.

### Cross-cutting

| Aspect | Result |
|---|---|
| Bugs / logic | none |
| Design / architecture | consistent with the established stocktake pattern (#242) |
| Security | Low, 0 vulnerabilities |
| Backward compatibility | retire is reversible via git; `.chezmoiremove` handles the deployed-copy cleanup |
| Tests | `make test` green (`1..257`); the 2 SHOULD items are missing-guard suggestions, not failures |

_No existing reviews to deduplicate._

tests/skill_provenance.bats
kryota-dev: **Deferred to a follow-up — valid finding, not addressed in this PR.**

This is a pre-existing asymmetry, not a regression introduced here: the retired-guard array mixes curated retirements (deployed from source, so they need a `.chezmoiremove` target) with unmanaged retirements that were never chezmoi-deployed (`patch-remote-control` / `find-skills` have no `.chezmoiremove` entry by design). A correct `.chezmoiremove`-existence assertion therefore has to split the array into those two groups rather than loop over all names — a larger refactor than this focused retire PR warrants. Tracking as follow-up work.

docs/agents/skills-provenance.md
kryota-dev: **Deferred to a follow-up — valid finding, not addressed in this PR.**

Also a pre-existing gap: `docs_facts.bats` verifies only the `FACT:curated-skill-count` marker against the directory count, not the per-theme `(N skills)` headings / backtick lists. An awk/grep test that checks each theme's count against its list length and that all themes sum to the FACT count is a sensible drift guard, but it is independent of this PR's scope (retiring one skill). Tracking as follow-up work.

